PR #1, opened 2018-02-25
Fix grammar in README.md

rcjsuen:
"it's" is a contraction of "it is". The following sentence doesn't make sense.

> Bash language server implementation based on Tree Sitter and it is grammar for Bash.

The possessive pronoun of "its" should be used instead in this context.


mads-hartmann:
@rcjsuen Woops, thanks!

PR #10, opened 2018-03-15
Fix typo in missing language server error

rcjsuen:
Should be `language` instead of `langauge`.


mads-hartmann:
@rcjsuen Thanks! :)

PR #14, opened 2018-03-30
Add Jest integration tests

skovhus:
To reduce the feedback loop of improving the language server and make contributions easier, I suggest that we add some Jest integration tests here.

It might have found an issue with the `wordAtPoint` function (see the FIXME).

If we like this, I think next step would be to set up CI. I can look into that if you want. : )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 
 vscode-client/out
 server/out
+.DS_Store
+*.log
+coverage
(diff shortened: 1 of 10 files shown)


mads-hartmann:
This is awesome @skovhus. I'm all for adding this to the CI as well. I know it might be a no-brainer but could I get you to write a tiny section in the README about how to run the tests?

skovhus:
I will. : )

skovhus:
@mads-hartmann let me know if the documentation looks good to you.

I also added a travis configuration here.

mads-hartmann:
@skovhus Thanks!

PR #15, opened 2018-04-01
Fix compilation

skovhus:
When introducing the integration tests (https://github.com/mads-hartmann/bash-language-server/pull/14) I broke the build. 

The TypeScript compilers output path depends magically on the files in the source directory. So when introducing new test setup files, the compilation path was suddenly `server/out/server/server.js` instead of `server/out/server.js`. 

The main problem here is that we don't clean the compilation directories... As I tested that it worked after introducing integration tests, but the old files were still in my local copy of `out`.

This PR:
- fixes the build
- introduces cleaning before building
- adds a convenience method for working on the server (compile and reinstall)

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -14,6 +14,9 @@
     "sourceMap": true
   },
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "**/__tests__/*",
+    "server/setupJest.ts",
+    "testing"
   ]
 }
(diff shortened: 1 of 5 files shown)


tsconfig.base.json
```diff
@@ -14,6 +14,9 @@
     "sourceMap": true
   },
   "exclude": [
-    "node_modules"
+    "node_modules",
+    "**/__tests__/*",
```
skovhus: By adding these the compilation path is as before `server/out/server.js`

skovhus:
FYI @mads-hartmann

mads-hartmann:
@skovhus Awesome, I love the `reinstall-server` improvement to the workflow 👍

PR #17, opened 2018-04-01
Completions based on your path

mads-hartmann:


mads-hartmann:
@skovhus I'd love if you could give this a review. I realise it's a bit messy, sorry about that :)

server/src/__tests__/executables.test.ts
```diff
@@ -0,0 +1,44 @@
...
+)
+
+describe('list', () => {
+  it('It finds executables on the PATH', () => {
+    expect.assertions(1)
+    const result = executables.then(ex => ex.list().find(x => x === 'iam-executable'))
+    return expect(result).resolves.toBeTruthy()
+  })
+
+  it.skip('It only considers files that have the executable bit set', () => {
```
skovhus: There is a skip here. 

Btw, you should remove the “It” from the message.

server/src/__tests__/executables.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect.assertions(1)
+    const result = executables.then(ex => ex.list().find(x => x === 'iam-not-executable'))
+    return expect(result).resolves.toBeFalsy()
+  })
+
+  it('It only considers executable directly on the PATH', () => {
+    expect.assertions(1)
+    const result = executables.then(ex =>
```
skovhus: Using `await` in these tests might make it easier. Then you don’t need the expect.assertions. : )

server/src/analyser.ts
```diff
@@ -1,207 +1,213 @@
...
-      namedDeclarations.push(
-        SymbolInformation.create(name, kinds[n.type], range(named), uri),
-      )
-      declarations[uri][name] = namedDeclarations
+export class Analyzer {
+  /**
+   * Initialize the Analyzer based on a connection to the client and an optional
+   * root path.
+   *
+   * If the rootPath is provided it will iniailize all *.sh files it can find
```
skovhus: Spelling: iniailize

server/src/analyser.ts
```diff
@@ -1,207 +1,213 @@
...
 type Texts = { [uri: string]: string }
-const texts: Texts = {}
 
 /**
- * Analyze the given document, cache the tree-sitter AST, and iterate over the
- * tree to find declarations.
- *
- * Returns all, if any, syntax errors that occurred while parsing the file.
- *
+ * The Analyzer uses the Abstract Syntax Trees (ASTs) that are provdied by
```
skovhus: Spelling: provdied

server/src/analyser.ts
```diff
@@ -1,207 +1,213 @@
...
-  const declarationsInFile = declarations[uri] || []
-  const ds = []
-  Object.keys(declarationsInFile).forEach(n => {
-    declarationsInFile[n].forEach(d => ds.push(d))
-  })
-  return ds
-}
+  // Global mapping from tree-sitter node type to vscode SymbolKind
+  private kinds: Kinds = {
+    environment_variable_assignment: LSP.SymbolKind.Variable,
```
skovhus: Any reason why we change to underscore here?

server/src/__tests__/executables.test.ts
mads-hartmann: @skovhus The `skip` is there because the test fails currently as it doesn't look at the executable bit yet  😅

mads-hartmann:
@skovhus I moved to using classes as I didn't like having state in the modules as top-level `const`s. With a class based approach it's clear what the state is and where it lives :) Are you against using classes? If so why? 😊

server/src/__tests__/executables.test.ts
```diff
@@ -1,44 +1,47 @@
...
 
-const executables = Executables.fromPath(
+const executablesPromise = Executables.fromPath(
   path.resolve(__dirname, '..', '..', '..', 'testing', 'executables'),
 )
 
 describe('list', () => {
-  it('It finds executables on the PATH', () => {
+  it('finds executables on the PATH', async () => {
     expect.assertions(1)
```
skovhus: Not that you are using `await` you don't need:
- `expect.assertions(1)`
- and returning the last `expect`

skovhus:
> With a class based approach it's clear what the state is and where it lives :) Are you against using classes? If so why?

So if you need state (which you do), then I usually go for small files with top-level state variable(s). The singleton module pattern makes this possible... If using classes makes this more clear then I'm all for it. 👍

server/src/__tests__/executables.test.ts
skovhus: @mads-hartmann this was the reason why I suggested using `await` here, so we could get rid of these extra things you need to remember to do. : )

server/src/analyser.ts
```diff
@@ -1,207 +1,214 @@
...
-    } else if (isDefinition(n)) {
-      const named = n.firstNamedChild
-      const name = contents.slice(named.startIndex, named.endIndex)
-      const namedDeclarations = declarations[uri][name] || []
-
-      namedDeclarations.push(
-        SymbolInformation.create(name, kinds[n.type], range(named), uri),
-      )
-      declarations[uri][name] = namedDeclarations
+export class Analyzer {
```
skovhus: I think we should an `export default class Analyzer` here, so import would be `import Analyzer from ...`.

server/src/executables.ts
```diff
@@ -0,0 +1,91 @@
...
+import * as Path from 'path'
+
+import * as ArrayUtil from './util/array'
+import * as FsUtil from './util/fs'
+import * as ShUtil from './util/sh'
+
+/**
+ * Provides information based on the programs on your PATH
+ */
+export class Executables {
```
skovhus: export default

server/src/executables.ts
```diff
@@ -0,0 +1,91 @@
...
+        : Promise.resolve(doc)
+    })
+  }
+}
+
+/**
+ * Find all the executables on the given path.
+ * Only returns direct children, or the path itself if it's an executable.
+ */
+function _executables(path: string): Promise<string[]> {
```
skovhus: We would call this `findExecutablesInPath`, then we can almost get rid of the code comment.

server/src/util/sh.ts
```diff
@@ -0,0 +1,25 @@
+import * as ChildProcess from 'child_process'
+
+/**
+ * Execute the following sh program.
```
skovhus: Nitpicking: I prefer not having code comments like this and make the function name more descriptive: `execShellScript(body: string)`

skovhus: We could use https://github.com/sindresorhus/execa for all child process handling. It has a few advantages but also introduces more dependencies. Pros cons.

server/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      .then(server => {
+        server.register(connection)
+        return server
+      })
+      .then(server => ({
+        capabilities: server.capabilities(),
+      }))
+  })
+
+  // Listen on the connection
```
skovhus: Nitpicking: superfluous comment

server/src/server.ts
```diff
@@ -1,173 +1,195 @@
...
+import * as LSP from 'vscode-languageserver'
+
+import { Analyzer } from './analyser'
+import { Executables } from './executables'
+
+/**
+ * The BashServer glues together the separate components to implement
+ * the various parts of the Language Server Protocol.
+ */
+export class BashServer {
```
skovhus: `export default`

server/src/analyser.ts
```diff
@@ -1,207 +1,214 @@
...
-      name = contents.slice(node.startIndex, node.endIndex)
-      rng = range(node)
-    } else if (isDefinition(n)) {
-      const namedNode = n.firstNamedChild
-      name = contents.slice(namedNode.startIndex, namedNode.endIndex)
-      rng = range(n.firstNamedChild)
-    }
+  // Global mapping from uri to the contents of the file at that uri.
+  // We need this to find the word at a given point etc.
+  private texts: Texts = {}
```
skovhus: We would rename all of these, so the comment would become superfluous.

Example: `uriToFileContent`, `uriToTreeSitterDocument`. I would prefer that. : )

server/src/util/sh.ts
mads-hartmann: @skovhus Let's switch to `execa` if we find bugs in this code :)

PR #21, opened 2018-04-07
Switch to yarn instead of npm

skovhus:
Try switching to `yarn` instead of `npm`.

This solves https://github.com/mads-hartmann/bash-language-server/issues/20


mads-hartmann:
@skovhus Thanks!

PR #40, opened 2018-04-26
More reliable npm command execution on windows

KonTrax:
Append '.cmd' extension when spawning child process for npm command on windows as '.cmd'/'.bat' files are not included when searching for executable by basename.


mads-hartmann:
Awesome, thanks @KonTrax 👍

Issue #18, opened 2018-04-06
Support new code folding api


mads-hartmann:
This is vscode specific, but should be very easy to add as we have the tree-sitter AST. https://code.visualstudio.com/updates/v1_22#_folding-provider-api

flying-sausages:
Never too late to bump an issue like this

skovhus:
@flying-sausages let me know if you would like to work on this. :)

flying-sausages:
I'd love to but I have 0 nodejs experience so I don't think I'll be of any help :(

2art:
This is no longer alive, is it?

loynoir:
Workaround

```jsonc
    "[shellscript]": {
      // workaround: zokugun.explicit-folding
      "explicitFolding.rules": [
        {
          // ref: /usr/bin/iptables-apply
          // ref: /etc/zsh/zshrc
          // case: `<< EOF `
          // case: `<< __EOF0__ `
          // case: `<< -EOF `
          "beginRegex": "<<\\s*[-'\"]*[A-Z0-9_]+['\"]*\\s*$",
          "endRegex": "^\\s*[A-Z0-9_]+$",
        },
      ]
    }
```

PR #71, opened 2018-08-02
Prototype implementation. Never seems to be invoked

mads-hartmann:
VSCode never seems to invoke the new folding ranges handler. Not
sure why.

Leaving it as is for now so I can investigate later.

### DIFF
--- a/server/src/analyser.ts
+++ b/server/src/analyser.ts
@@ -287,6 +287,30 @@ export default class Analyzer {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


mads-hartmann:
This is not ready to be merged but I'm leaving it open so I won't forget.

mads-hartmann:
This is to solve #18 by the way

mads-hartmann:
It has been a few years, I think we can safely close this ;)

PR #78, opened 2018-08-21
Add missing nodes warning + grammar update

skovhus:
Closes https://github.com/mads-hartmann/bash-language-server/issues/74

Currently, the extension is just silent if a node is missing, but this PR adds an error to the bottom of a file if a node is missing.

screenshot 2018-08-21 22 05 07

Unfortunately, we cannot show the missing node where it occurs... So we should evaluate if this PR gives any additional value. See https://github.com/tree-sitter/tree-sitter-bash/issues/30

Besides, I'm bumping the version number and increase the required server version to the current release (a lot are probably using an old version).

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bash Language Server
 
+## 1.5.3
+
+* Support for showing warning for missing nodes
+
 ## 1.5.2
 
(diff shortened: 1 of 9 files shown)


mads-hartmann:
@skovhus I'll merge and release tomorrow 👍

skovhus:
I've sneaked in a small upgrade of the grammar in this PR. 😊

PR #113, opened 2019-03-23
[WIP] More file extensions than .sh and shebang lookup

skovhus:
This allows more file extensions than just .sh. Besides, we limit the files being analyzed to the ones that have a `bash` / `sh` shebang.

While doing this, I was able to remove some of the `fs.existsSync` and `fs.lstatSync` lookups, that would be done directly in `glob`. 

I also tried `fast-glob`, but the matching algorithm didn't seem that stable (although a lot faster).

To support file names without file extensions, we probably need another glob `glob.sync('**/!(*.*)')`.

Related: https://github.com/mads-hartmann/bash-language-server/issues/47

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -17,7 +17,7 @@
     "node": "*"
   },
   "dependencies": {
-    "glob": "^7.1.2",
+    "glob": "^7.1.3",
     "request": "^2.83.0",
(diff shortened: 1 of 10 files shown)


skovhus:
Replaced by https://github.com/mads-hartmann/bash-language-server/pull/187

PR #119, opened 2019-04-08
Fix issue #83

### LINKED ISSUE #83: VSCode prompts non-sense.
I am using Bash IDE extension in VSCode. And installed bash-language-server as descripted in its details.

Now I have two problems. One is apparently something is wrong. The other is "I am not sure".

1. Code prompt are all non-sense. For example, I typed "then", I got "thinkjettopbm". I typed "case", I got "catchsegv".

2. Code completion does not work. In my idea, this is something like, when I typed "case", it would insert the "esac". But in my IDE, it does not. Just marks a bunch of lines "syntax error" red underlines.

What should I do?

thomasjm:
This is an attempt to fix #83. See my comments there for more context on the problem.

This PR adds filtering to the completion results, so you only get complete results for which the current word you're typing is a prefix.

The main challenge here was to improve the `analyser.wordAtPoint` function so that it retrieves the correct word when you're at the edge of a word (as is usually the case when using completion).

Note: I suspect that similar boundary issues are the reason for the "Hack" code on line 321 of the original code, so maybe that can be removed now but I'm not sure how to test.

@skovhus

### DIFF
--- a/server/src/analyser.ts
+++ b/server/src/analyser.ts
@@ -312,18 +312,85 @@ export default class Analyzer {
     const document = this.uriToTreeSitterTrees[uri]
     const contents = this.uriToFileContent[uri]
 
-    const node = document.rootNode.namedDescendantForPosition({ row: line, column })
+    if (!document.rootNode) {
+      // Check for lacking rootNode (due to failed parse?)
+      return null
+    }
+
+    const point = { row: line, column }
+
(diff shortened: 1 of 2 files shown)


mads-hartmann:
Thanks a lot for creating this PR @thomasjm. If you could include a test or two that would be lovely.

skovhus:
Released as part of https://github.com/mads-hartmann/bash-language-server/pull/192

Thanks for your contribution!

PR #132, opened 2019-05-29
Add compopt bash builtin command.

akurtakov:
Signed-off-by: Alexander Kurtakov <akurtako@redhat.com>


skovhus:
Thanks. `compopt` is actually not a built-in on OS X, but I see that it is usually there on other installations.

PR #142, opened 2019-08-16
Improve error discovery when the LSP process cannot be started

robertrossmann:
When the `bash-language-server` cannot be started or found the user only sees a message that the process could not be found instead of seeing the actual reason for the failure.

This message got me very confused when I installed the extension because I made sure I had the `bash-language-server` installed and could find it in my `PATH` but I thought VS Code somehow fails to find it. The actual problem was that I upgraded to Node.js 12 and the process failed to start as it was not compiled against the correct Node.js version. I only discovered that when I actually tried to start the server.

This PR improves the discovery of the failure by showing a more generic error message and guiding users to the DevTools console where the actual error is logged.

### DIFF
--- a/vscode-client/src/extension.ts
+++ b/vscode-client/src/extension.ts
@@ -29,7 +29,7 @@ export async function activate(context: ExtensionContext) {
 
     start(context, command, explainshellEndpoint, highlightParsingErrors)
   } catch (error) {
-    handleMissingExecutable()
+    handleStartError(error)
   }
 }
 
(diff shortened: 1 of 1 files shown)


skovhus:
Thanks! Can you run a `yarn run lint` and amend the formatting changes?

robertrossmann:
Hi @skovhus, fixed and force-pushed! 🎉

I also update the `console.log` statement to `console.error` to show the problem in red in the console, since it's really an error if the extension is unable to start.

HsunGong:
That's good, is it showing error info in `output` or `debug-console` ? May change more info in this hint.

robertrossmann:
@HsunGong the full stack trace and error message will be visible in the DevTools console in VS Code.

PR #145, opened 2019-09-16
 see More error info in vscode `extension.js`

HsunGong:
To see More error info in vscode.
Why the server is not running which may help solve issues like
https://github.com/mads-hartmann/bash-language-server/issues/41
https://github.com/mads-hartmann/bash-language-server/issues/96

Please Output More info if needed.

By the way, may and how can i use `base-language-server` in shells or web?


skovhus:
Thanks for contributing! This was also solved in https://github.com/mads-hartmann/bash-language-server/pull/142

skovhus:
> By the way, may and how can i use base-language-server in shells or web?

You need a client that implements the LSP protocol. https://langserver.org/

Issue #47, opened 2018-06-07
More file extensions than just "*.sh"?


nwinkler:
It would be great if the user could provide more file extensions to parse. At the moment, only `*.sh` files are found and parsed:

https://github.com/mads-hartmann/bash-language-server/blob/00941c001ca342c8a9c1b647b2a1b7ea703e5fc6/server/src/analyser.ts#L47

When working with Bash scripts that use a different extension, e.g. `.bash`, the bash language server currently does not work.

It would be great if there was a configuration option that would allow to list the extensions that should be scanned...

mads-hartmann:
@nwinkler This is a great idea. I'd be more than happy to merge a PR adding this feature 👍

nwinkler:
It's been a while since I've done any Node.js/TypeScript, and I've never worked on an extension for VSCode, but that shouldn't stop me...

nwinkler:
Found the developer guide - I'll take a look at that and see how far I can get. More time in the day would be nice...

mads-hartmann:
@nwinkler Awesome 👏 Let me know if you have any questions

Lucretia:
I also use .inc for include files. I'm not the only person who does this.

bunam:
add .command on mac ?

i think a better solution is to look in the 1st line for began by '#!' and containing 'bash' 'sh' ... ?
#!/usr/bin/env bash

maximbaz:
Or rely on mime type, as that one already respects shebang:

```
❯ file --mime-type $(readlink -m ~/bin/backup)
text/x-shellscript
```

skovhus:
Good point with the symlinks @maximbaz

Any PRs would be much appreciated.

skovhus:
So we have two use cases:
1) files without .sh extensions that are bash scripts
2) file with .sh extensions that are not bash scripts

Suggested solution: **Read the shebang line for all potential shell script files** (.sh, .command, .inc, and no file extension).

As a nice to have on top, we should make the glob pattern configurable.

przepompownia:
@skovhus after merging the current master into your branch I looked if and how it works (see #156 ). 

The essence of your change is that you make an **intersection** of having shebang **and** matching the glob pattern. It is not what I expect. I would like to use an **alternative** of these to determine the type. In such case we probably need to scan `rootPath` twice: first against the custom glob pattern, then against shebang, with another (less restricted) glob pattern.

skovhus:
[...]

PR #156, opened 2019-11-06
Shebang handling

przepompownia:


przepompownia:
At the moment this PR merges only the current master into `shebang-handling`. 
See #47

skovhus:
A bit hard to read this PR. Could we rebase the changes on top of master instead of merging?

przepompownia:
TODO
- [ ] review (I am beginner with TS-specific things) 
- [ ] allow to pass glob pattern in another way than using an environment variable
- [ ] more tests
- [ ] allow the user to disable scanning working dir against bash/sh shebang
- [ ] give `hasBashShebang` only the first line of the file (do not match the entire content)

skovhus:
Sorry for stalling here. I've investigated the issue, and it turns out that all bash files (that VS Code can detect) are handled just fine. We don't need to read all the files in a project–unless we want to do just to definition across files (required more work).

Screenshot 2020-02-27 10 41 54
Screenshot 2020-02-27 10 42 01

przepompownia:
It seemed to be natural that the results of `analyze`ing the rest of scripts are accessible for completion and jumping. Indeed: at the moment there is no need to make (almost?) unusable index.

skovhus:
> It seemed to be natural that the results of analyzeing the rest of scripts are accessible for completion and jumping. Indeed: at the moment there is no need to make (almost?) unusable index.

My mistake. It is indeed used for indexing. https://github.com/mads-hartmann/bash-language-server/issues/47#issuecomment-593313329 👍

skovhus:
@przepompownia thanks for your work here. We now handle most of the cases here – except for reading all files. 

I'm reluctant to implement parsing all the files in a project, that shouldn't be necessary. We might consider adding parsing all files without extensions though. :)

Issue #155, opened 2019-11-05
Completion of exacutables in PATH with tilde


dmerc:
Given `PATH=/usr/bin:~/bin` only the executables in `/usr/bin` are shown in auto completion.

I found that its due to problem in executables.ts, function findExecutablesInPath,
the call to fs.lstat doesnt expand tilde in the path.

I suggest just adding tilde expansion, since nodejs doesnt handle it automatically

skovhus:
That makes sense, thanks. Do you want to give it a try in a PR?

dmerc:
ok will do

PR #157, opened 2019-11-11
add PATH tilde expansion

dmerc:
implement #155

### DIFF
--- a/server/src/util/fs.ts
+++ b/server/src/util/fs.ts
@@ -1,4 +1,5 @@
 import * as Fs from 'fs'
+import * as Os from 'os'
 
 export function getStats(path: string): Promise<Fs.Stats> {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


server/src/util/fs.ts
```diff
@@ -11,3 +12,10 @@ export function getStats(path: string): Promise<Fs.Stats> {
     })
   })
 }
+
+export function untildify(pathWithTilde: string): string {
+  const homeDirectory = Os.homedir()
```
skovhus: I would a comment with a reference to where this comes from. I assume https://github.com/sindresorhus/untildify/blob/master/index.js#L11

dmerc:
yes, that code came from sindresorhus/untildify.
I added the comment, amended last commit. Please review.

server/src/util/fs.ts
```diff
@@ -11,3 +12,11 @@ export function getStats(path: string): Promise<Fs.Stats> {
     })
   })
 }
+
+// from https://github.com/sindresorhus/untildify/blob/master/index.js#L11
```
rcjsuen:
```suggestion
// from https://github.com/sindresorhus/untildify/blob/f85a087418aeaa2beb56fe2684fe3b64fc8c588d/index.js#L11
```
Should use the canonical form instead because `master` branch is always changing.

PR #171, opened 2020-01-06
Update README.md: add JupyterLab client

krassowski:
Thank you for maintaining this language server! Just wanted to let you know that we support bash using your awesome package, and add a shameless promotion to help users who may look for a client with bash support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ The following editors and IDEs have available clients:
 - [Oni](https://github.com/onivim/oni) ([see below](#oni))
 - Eclipse ([ShellWax](https://marketplace.eclipse.org/content/shellwax))
 - Emacs ([see below](#emacs))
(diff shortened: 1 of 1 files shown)


krassowski:
Yes! In three ways:
1. As a standalone kernel for notebooks after installation of the [bash kernel](https://github.com/takluyver/bash_kernel):
   
   bash_notebook

2. To edit bash files (JupyterLab has not only the notebook interface but also a normal editor):
   
   bash_file

3. Embedded in IPython kernel using magics (`%%`) and shell expressions (`!`):
   
   Screenshot from 2020-01-07 10-12-22

The last one is not hooked with the LSP yet but will be by the next release, as I mostly use bash as (3) and (2).

skovhus:
Very nice! Thanks

PR #187, opened 2020-03-02
Extend and make glob pattern configurable

skovhus:
### Context

Issue: https://github.com/mads-hartmann/bash-language-server/issues/47

This builds on the work done a year ago (https://github.com/mads-hartmann/bash-language-server/pull/113). I ended up re-implementing it to get back into the problem space. Thanks to @przepompownia for re-booting this (https://github.com/mads-hartmann/bash-language-server/pull/156).

### This PR

Here we add support for configuring the glob pattern used when pre-analyzing all bash files (this is done to support jump to definition). Besides making it configurable we extend it from `**/*.sh` to `**/*@(.sh|.inc|.bash|.command)`.

I recommend reviewing commit by commit. 

### Next steps

This will introduce a few false positives as we do not parse the shebang or mime types  (yet), as I tried to document in the fixture folder.

Next step might be to implement the shebang part of https://github.com/mads-hartmann/bash-language-server/pull/113 or mime types (https://github.com/mads-hartmann/bash-language-server/issues/47#issuecomment-409279032) and consider the ideas introduced in https://github.com/mads-hartmann/bash-language-server/pull/156

But the challenge is that not everyone use shebangs in their bash files... so I guess we shouldn’t make that a filter? 

If we can get the included files out of tree sitter, then we could potentially parse those files.

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Bash Language Server
 
+## 1.8.0
+
+* Extend file glob used for pre-analyzing files from `**/*.sh` to `**/*@(.sh|.inc|.bash|.command)`
+* Make file glob configurable with `GLOB_PATTERN` environment variable
+
 ## 1.7.0
 
 * Add PATH tilde expansion
--- a/server/package.json
+++ b/server/package.json
@@ -18,7 +18,7 @@
     "node": "*"
   },
   "dependencies": {
-    "glob": "^7.1.2",
+    "glob": "^7.1.6",
     "request": "^2.83.0",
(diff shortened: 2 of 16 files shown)


skovhus:
A few stats to guide upcoming work (shebang/mimetype handling):

Globbing files inside our `node_modules` (just to have a bunch of files):
1) `**/*@(.sh|.inc|.bash|.command)` and `**/*.sh` resolves in 3 seconds (with 3 files)
2) `**/*` resolves in 5 seconds with 14K files

Reading the file content (sync) of the 14K files takes around 6 seconds. This would need to be done if we analyze the shebang of all files.

Issue #197, opened 2020-03-04
How to enable workspace symbol list support?


przepompownia:
First I ensured that I use the latest version ` "message": "Initialized server v. 1.11.0 for file:///home/***/.githooks"` ;) The I checked the capability using `CocHasProvider`:
```
echo CocHasProvider('workspaceSymbols')
v:false
CocList symbols
[coc.nvim] Error on "CocList symbols": Workspace symbols provider not found for current document
```
However, trying to execute `call CocRequest('bash','workspace/symbol', {'query': 'single'})` ends with no error but also no results (not as expected):
```
[Trace - 16:47:04] Sending request 'workspace/symbol - (4)'.
Params: {
    "query": "single"
}


[Trace - 16:47:04] Received notification 'window/logMessage'.
Params: {
    "type": 4,
    "message": "onWorkspaceSymbol"
}


onWorkspaceSymbol
[Trace - 16:47:04] Received response 'workspace/symbol - (4)' in 3ms.
Result: []
```
(at least one locally defined function was expected). 

Is the request correct? I looked at https://microsoft.github.io/language-server-protocol/specification#workspace_symbol

skovhus:
Can you verify that other symbols doesn’t work either? Maybe paste a small example here of the shell script?

skovhus:
> Is matching to the head of symbol name required by LSP or is it your choice? 

That was just what was picked when implementing. It makes sense that the search is using includes instead of startsWith. I would like to avoid making it configurable and hope that includes works for everyone. It also looks like this is what other LSP servers are doing.

Does https://github.com/mads-hartmann/bash-language-server/pull/198 work for you?

przepompownia:
It works and make searching much more useful (assuming that the server will correctly inform the client about existing this functionality) :+1: 
However I still vote for make it like at the example below (tsserver used with CoC):
![camel](https://user-images.githubusercontent.com/11404453/75915162-4ec5b680-5e56-11ea-9b61-c0fea9342b27.png)

skovhus:
@przepompownia thanks, I'll look into fuzzy searching. 

The provider/discoverability issue: it is called `workspaceSymbolProvider` in capabilities. Maybe you should write `workspaceSymbol` (not plural)?

przepompownia:
```
Error on request (hasProvider): Vim(return):Error invoking 'hasProvider' on channel 3 (coc):                                                                                  
workspaceSymbol not supported.                                                                                                                                                
0     
```

[...]

PR #198, opened 2020-03-04
Fuzzy search workspace symbols 

skovhus:
Related https://github.com/mads-hartmann/bash-language-server/issues/197

@przepompownia rightfully identified that workspace symbols are more useful if they are resolved using fuzzy search.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Please see [docs/development-guide][dev-guide] for more information.
 [tree-sitter]: https://github.com/tree-sitter/tree-sitter
 [tree-sitter-bash]: https://github.com/tree-sitter/tree-sitter-bash
 [vscode-marketplace]: https://marketplace.visualstudio.com/items?itemName=mads-hartmann.bash-ide-vscode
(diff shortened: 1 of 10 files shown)


przepompownia:
Please do not automatically close #197 by resolving only the matching problem. That issue contains also the problem of discoverability by the client.

skovhus:
@przepompownia I've updated this to use fuzzy search. Please try it out if you have time. :)

przepompownia:
I'm trying to take advantage of the fact that we have the opportunity to work at the same time.

This request takes lot of CPU usage and does not want to end. It looks that I have to manually kill node process :) The next step will be trying to preview what happens.

przepompownia:
It was happen without fetching the new package. I only compiled the changes fom this PR.

przepompownia:
It works after fetching the new dependency :)

I have added primitive git hooks in .githooks (with `core.hooksPath`) that runs `yarn` and `yarn run compile` but I had to run them manually after getting forced update.

PR #212, opened 2020-04-14
replace readFileSync

Yanpas:
I decided to do it sequentially to keep the same behaviour and predictable memory consumption.

### DIFF
--- a/server/src/analyser.ts
+++ b/server/src/analyser.ts
@@ -1,4 +1,4 @@
-import * as fs from 'fs'
+import {promises as fs} from 'fs'
 import * as FuzzySearch from 'fuzzy-search'
 import * as request from 'request-promise-native'
 import * as URI from 'urijs'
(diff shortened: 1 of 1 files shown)


skovhus:
I realize now that using the `fs.promises` API means that we limit the support to node 10+.

So either we should:
1) Update the required node version here https://github.com/bash-lsp/bash-language-server/blob/master/server/package.json#L18
2) Use the `util.promisify()` API (node 8) as you originally suggested

Thoughts?

Yanpas:
I think the second one is better. Ubuntu 18.04 ships 8th lts version e.g.

skovhus:
Let us go with 2 then. 👍 I’ll update CI to test on node 8.

skovhus:
Replaced by https://github.com/bash-lsp/bash-language-server/pull/213

Issue #223, opened 2020-04-16
EACCES error on directory w/o permissoins


jordemort:
Similar to https://github.com/bash-lsp/bash-language-server/issues/202

I'm using VS Code's Remote SSH functionality to connect to a development server that keeps my workspace mounted on its own file system. Unfortunately, that means my workspace contains a `lost+found` directory that I don't have permission to read. This leads to the extension failing to start with:

```
Request initialize failed with message: EACCES: permission denied, scandir '/workspace/enterprise2/lost+found'
```

miwarin:
I also encountered EACCESS. I want you to ignore EACCESS and continue scanning.

In my case, I could not access because the directory to be scanned is owned by root:root.

```
[Info  - 13:02:32] BashLanguageServer initializing...
Analyzing files matching glob "**/*@(.sh|.inc|.bash|.command)" inside /home/miwa/work/xxxx
glob error { [Error: EACCES: permission denied, scandir '/home/miwa/work/xxxx/root']
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path:
   '/home/miwa/work/xxxx/root' }
[Error - 13:02:33] Server initialization failed.
  Message: Request initialize failed with message: EACCES: permission denied, scandir '/home/miwa/work/xxxx/root'
  Code: -32603 
```


```shell
miwa@archlinux[~/work/xxxx]% sudo ls -l root
total 40
drwxr-xr-x 2 root root 36864 Apr 16 15:26 aaaa
drwxr-xr-x 2 root root  4096 Apr 16 15:26 bbbb
```

skovhus:
Thanks for reporting this. https://github.com/bash-lsp/bash-language-server/pull/224 fixes the issue by not crashing the server but just skipping cross-file references. There is also a warning encouraging you to update the glob or fix permission issues.

Hope this solves the issue for you.

jordemort:
I'm not sure how to update the glob in the VS Code extension to exclude that directory; I tried excluding it from the editor using `.vscode/settings.json` and that removes it from view from VS Code's perspective but the Bash IDE extension is still trying and failing to scan the directory. There is definitely nothing in there that other scripts reference.

skovhus:
@jordemort fair point. I figured out a way to handle this case by letting the globbing library run in non-strict mode.

skovhus:
A fix has been released, a new server and vscode client is out. 

Let me know if it works for you.

jordemort:
[...]

PR #226, opened 2020-04-18
Glob in non-strict mode

skovhus:
Run glob in non-strict mode to recover from errors (e.g. permission issues).

Originally, it looked like we were already running in non-strict mode. But the [documentation](https://github.com/isaacs/node-glob/#options) for node-glob could be made more clear (https://github.com/isaacs/node-glob/issues/298).

Related: https://github.com/bash-lsp/bash-language-server/pull/224 #223 https://github.com/isaacs/node-glob/pull/413


skovhus:
Upstream fix: https://github.com/isaacs/node-glob/pull/413

PR #231, opened 2020-05-01
Support 0/1 as value for HIGHLIGHT_PARSING_ERRORS

damienrg:
We permit to specify 0 or 1 as value for HIGHLIGHT_PARSING_ERRORS
because the variable has a boolean type and because it is common to use
these values for feature toggle in environment variables.


skovhus:
@damienrg thanks for your contribution. It makes sense to support both cases.

skovhus:
This was released as bash-language-server@1.14.0.

PR #234, opened 2020-05-14
Feature comment docs onhover

nikita-skobov:
In most languages, comments above a function serve as documentation for that function. I modified this extension to be able to view the comments that are above functions as tooltips onHover.

It looks like this:

```sh
# this will not be included because
# there is an empty line below

# single positional argument: the commit Id
# returns true if that commit id
# is a merge commit
is_merge_commit() {
    # does something based on $1
    if [[ $1 == "something" ]]; then
        return 0
    fi
}
```

defined_this_file

Also works for functions defined in other files that are included in the `this.uriToDeclarations` from analyser.ts:

```sh
# fn_exists.sh

# does a function exist in this script?
# args: $1 the name of the function
fn_exists() {
    if [[ $1 == "exists" ]]; then
        return 0
    fi
}
# end of fn_exists.sh


# main_script.sh
fn_exists "my_func"
```
defined_other_file

I followed the steps in [the development guide](https://github.com/bash-lsp/bash-language-server/blob/master/docs/development-guide.md) before editing the extension and I got this failure:

```
 FAIL  server/src/__tests__/server.test.ts
  ● server › responds to onCompletion with filtered list when word is found
```

so that error is unrelated to my changes. I then ran `yarn test` again after making my changes, and no other tests failed.

I'm new to working on VScode extensions, so let me know if there is a more idiomatic/efficient way to achieve what my code does.

### DIFF
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -195,10 +195,10 @@ describe('findSymbolCompletions', () => {
 describe('commentsAbove', () => {
   it('returns a string of a comment block above a line', () => {
     analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
-    expect(analyzer.commentsAbove(CURRENT_URI, 22)).toEqual(' doc for func_one')
+    expect(analyzer.commentsAbove(CURRENT_URI, 22)).toEqual('doc for func_one')
 
     expect(analyzer.commentsAbove(CURRENT_URI, 28)).toEqual(
-      ' doc for func_two\n has two lines',
+      'doc for func_two\nhas two lines',
     )
 
     // if there is a line break in the comments
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -192,6 +192,21 @@ describe('findSymbolCompletions', () => {
   })
 })
 
+describe('commentsAbove', () => {
+  it('returns a string of a comment block above a line', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
+    expect(analyzer.commentsAbove(CURRENT_URI, 22)).toEqual(' doc for func_one')
+
+    expect(analyzer.commentsAbove(CURRENT_URI, 28)).toEqual(
+      ' doc for func_two\n has two lines',
+    )
+
+    // if there is a line break in the comments
+    // it should not include the above comment
+    expect(analyzer.commentsAbove(CURRENT_URI, 36)).not.toMatch('this is not included')
+  })
+})
+
 describe('fromRoot', () => {
   it('initializes an analyzer from a root', async () => {
     const parser = await initializeParser()
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -196,14 +196,28 @@ describe('commentsAbove', () => {
   it('returns a string of a comment block above a line', () => {
     analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
     expect(analyzer.commentsAbove(CURRENT_URI, 22)).toEqual('doc for func_one')
+  })
 
+  it('handles line breaks in comments', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
     expect(analyzer.commentsAbove(CURRENT_URI, 28)).toEqual(
-      'doc for func_two\nhas two lines',
+        'doc for func_two\nhas two lines',
     )
+  })
 
-    // if there is a line break in the comments
-    // it should not include the above comment
-    expect(analyzer.commentsAbove(CURRENT_URI, 36)).not.toMatch('this is not included')
+  it('only returns connected comments', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
+    expect(analyzer.commentsAbove(CURRENT_URI, 36)).toEqual('doc for func_three')
+  })
+
+  it('returns null if no comment found', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
+    expect(analyzer.commentsAbove(CURRENT_URI, 45)).toEqual(null)
+  })
+
+  it('works for variables', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
+    expect(analyzer.commentsAbove(CURRENT_URI, 42)).toEqual('works for variables')
   })
 })
 
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -201,7 +201,7 @@ describe('commentsAbove', () => {
   it('handles line breaks in comments', () => {
     analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
     expect(analyzer.commentsAbove(CURRENT_URI, 28)).toEqual(
-        'doc for func_two\nhas two lines',
+      'doc for func_two\nhas two lines',
     )
   })
 
@@ -239,8 +239,7 @@ describe('fromRoot', () => {
 
     expect(connection.window.showWarningMessage).not.toHaveBeenCalled()
 
-    // TODO: maybe not hardcode this number in case
-    // fixture files are added in the future?
+    // if you add a .sh file to testing/fixtures, update this value
     const FIXTURE_FILES_MATCHING_GLOB = 11
 
     // Intro, stats on glob, one file skipped due to shebang, and outro
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -210,7 +210,9 @@ describe('fromRoot', () => {
 
     expect(connection.window.showWarningMessage).not.toHaveBeenCalled()
 
-    const FIXTURE_FILES_MATCHING_GLOB = 10
+    // TODO: maybe not hardcode this number in case
+    // fixture files are added in the future?
+    const FIXTURE_FILES_MATCHING_GLOB = 11
 
     // Intro, stats on glob, one file skipped due to shebang, and outro
     const LOG_LINES = FIXTURE_FILES_MATCHING_GLOB + 4
--- a/server/src/analyser.ts
+++ b/server/src/analyser.ts
@@ -399,6 +399,66 @@ export default class Analyzer {
     return name
   }
 
+  /**
+   * Find a block of comments above a line position
+   */
+  public commentsAbove(uri: string, line: number): string | null {
+    const doc = this.uriToTextDocument[uri]
+
+    const comment_block = []
+
+    // start from the line above
+    let comment_block_index = line - 1
+
+    // check if that line starts with a comment
+    // its possible that some lines
+    // might have whitespace before the comment
+    // begins, so check by iterating from
+    // the start of the line
+    const starts_with_a_comment = (line: string): boolean => {
+      let char_index = 0
+      let char = line.charAt(char_index)
+      while (char === ' ') {
+        char_index += 1
+        char = line.charAt(char_index)
+      }
+      // once we reach a character that is not whitespace
+      // return true if its a comment, false otherwise
+      return char === '#'
+    }
+
+    const exists_and_is_comment = (elm: string): boolean => elm !== undefined && starts_with_a_comment(elm)
+
+    let current_line = doc.getText({
+      start: { line: comment_block_index, character: 0 },
+      end: { line: comment_block_index + 1, character: 0 },
+    })
+
+    while (exists_and_is_comment(current_line)) {
+      // TODO: vscode must have some API for detecting comments
+      // should figure that out instead of hardcoding the # symbol...
+      current_line = current_line.replace('#', '')
+      if (current_line.charAt(current_line.length - 1) === '\n') {
+        // TODO: should preserve comment line endings?
+        // I think not because sometimes comment strings wrap a line
+        // to prevent the line from being too long, and it'd be nice
+        // to see it displayed as a sentence.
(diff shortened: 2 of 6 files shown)


skovhus:
I’m curious to know why the tests are failing locally for you. As you see CI isn’t failing, all tests pass. So it must be something in your environment that the test suite is not catering for. Can you share more information/output from the failure. Is there more output from when running the tests? Which OS are you running?

nikita-skobov:
I'm a little confused on the architecture with the server. Before I run the tests do I need to do something to set up the server? In any case here is my system information:

```
OS/distro: Arch linux
node --version: v13.12.0
npm --version: 6.14.4
yarn --version: 1.22.4
vscodium --version: 1.45.0 x64
(I use vscodium instead of vscode maybe that has something to do with it)
```

I recloned the repo in a new directory straight from master, and I did the following steps:

```
git clone https://github.com/bash-lsp/bash-language-server

cd bash-language-server

yarn install
# output from yarn install:
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.11: The platform "linux" is incompatible with this module.
info "fsevents@1.2.11" is an optional dependency and failed compatibility check. Excluding it from installation.
info fsevents@2.1.2: The platform "linux" is incompatible with this module.
info "fsevents@2.1.2" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...

$ cd server && yarn install && cd ../vscode-client && yarn install && cd ..
yarn install v1.22.4
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...

Done in 1.40s.
yarn install v1.22.4
[1/5] Validating package.json...
warning bash-ide-vscode@1.9.1: The engine "vscode" appears to be invalid.
[2/5] Resolving packages...
[3/5] Fetching packages...
warning vscode-languageclient@6.1.3: The engine "vscode" appears to be invalid.
[4/5] Linking dependencies...
[5/5] Building fresh packages...
Done in 1.24s.
Done in 9.04s.

yarn run compile
# output from yarn run compile:
yarn run v1.22.4
$ tsc -b
Done in 9.76s.

yarn run lint
# output from yarn run lint:
yarn run v1.22.4
$ yarn lint:bail --fix
$ eslint . --ext js,ts,tsx --cache --fix
Done in 4.11s.

yarn run test
# output from yarn run test:
yarn run v1.22.4
$ jest --runInBand
 PASS  server/src/util/__tests__/sh.test.ts
 FAIL  server/src/__tests__/server.test.ts
  ● server › responds to onCompletion with filtered list when word is found

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      226 | 
      227 |     // Limited set (not using snapshot due to different executables on CI and locally)
    > 228 |     expect(result && 'length' in result && result.length < 5).toBe(true)
          |                                                               ^
      229 |     expect(result).toEqual(
      230 |       expect.arrayContaining([
      231 |         {

      at server/src/__tests__/server.test.ts:228:63
      at fulfilled (server/src/__tests__/server.test.ts:5:58)

 PASS  server/src/__tests__/analyzer.test.ts
 PASS  server/src/__tests__/config.test.ts
 PASS  server/src/__tests__/executables.test.ts
 PASS  server/src/util/__tests__/shebang.test.ts
 PASS  server/src/util/__tests__/flatten.test.ts
 PASS  server/src/util/__tests__/array.test.ts

Test Suites: 1 failed, 7 passed, 8 total
Tests:       1 failed, 1 skipped, 52 passed, 54 total
Snapshots:   18 passed, 18 total
Time:        7.746s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I hope that helps. Let me know if theres any other info needed

nikita-skobov:
In regards to the feature PR, I will try my hand at writing some tests. I suppose the tests I need to write/modify are:

- modify the onHover test: https://github.com/nikita-skobov/bash-language-server/blob/68a15ff3c7a578ecf951e1c5754065d745020b67/server/src/__tests__/server.test.ts#L50
- add a new unit test for the function: `commentsAbove` in `analyser.ts`

I can see (somewhat) how to modify the existing test to test for the comment doc feature, but Im not sure I'd know how to go about creating a new test for the `commentsAbove` function.

In regards to the AST, can you give me some suggestions on how to use it to extract comments at a specific line?

skovhus:
Can I get you to do a `console.log(result)` just before the failing assertion and paste the result here. But it is 99% something in your local environment that we are not catering for. 

As for the tests, then I can come with some suggestions for these tests if you are fine with that. We basically want to test the different cases (one comment above, multiple comments, no comments) and see how the function reacts to this. After that we can refactor the function and be sure it still works. 

About the AST, then it currently doesn’t support this (I investigated it). So your current approach is fine.

nikita-skobov:
Yes. Here is the output:

```
yarn run v1.22.4
$ jest --runInBand
 FAIL  server/src/__tests__/server.test.ts
  ● Console

    console.log server/src/__tests__/server.test.ts:227
      [
        { label: 'rm', kind: 12, data: { name: 'rm', type: 1 } },
        { label: 'rm_pulse', kind: 12, data: { name: 'rm_pulse', type: 1 } },
        { label: 'rmdir', kind: 12, data: { name: 'rmdir', type: 1 } },
        { label: 'rmid', kind: 12, data: { name: 'rmid', type: 1 } },
        {
          label: 'rmiregistry',
          kind: 12,
          data: { name: 'rmiregistry', type: 1 }
        }
      ]

  ● server › responds to onCompletion with filtered list when word is found

    expect(received).toBe(expected) // Object.is equality

    Expected: true
    Received: false

      228 | 
      229 |     // Limited set (not using snapshot due to different executables on CI and locally)
    > 230 |     expect(result && 'length' in result && result.length < 5).toBe(true)
          |                                                               ^
      231 |     expect(result).toEqual(
      232 |       expect.arrayContaining([
      233 |         {

      at server/src/__tests__/server.test.ts:230:63
      at fulfilled (server/src/__tests__/server.test.ts:5:58)

 PASS  server/src/util/__tests__/sh.test.ts
 PASS  server/src/__tests__/analyzer.test.ts
 PASS  server/src/__tests__/config.test.ts
 PASS  server/src/__tests__/executables.test.ts
 PASS  server/src/util/__tests__/shebang.test.ts
 PASS  server/src/util/__tests__/flatten.test.ts
 PASS  server/src/util/__tests__/array.test.ts

Test Suites: 1 failed, 7 passed, 8 total
Tests:       1 failed, 1 skipped, 52 passed, 54 total
Snapshots:   18 passed, 18 total
Time:        6.37s, estimated 7s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Maybe Im wrong but it seems the issue is that the test assumes that there are less than 5 executables on a user's PATH that begin with `rm`, whereas my system has 5. so the `result.length < 5` expectation fails

Solution: update to less than 6? Or otherwise it probably doesn't matter much because this test fails based on a user's path, and as long as it works on the CI, does it really matter if a user has a bunch of extra executables that start with `rm`?

nikita-skobov:
I added the tests mentioned above using a new fixture: `comment-doc-on-hover.sh` which has some different examples of comments being shown on the tooltip.

Also, I wonder whats the best way to display the comments, ie: should it preserve newlines or not? Should it attempt to style it differently somehow?

The current functionality will not preserve comment newlines, eg:

```
# this is
# a comment
my_func() { }
```

will be returned from `commentsAbove` as:
```
 this is\n a comment
```

but because it's rendered with markdown it will show up as:

```
 this is a comment
```

Personally I think this is fine because comment documentations are typically wrapped around when the line gets too long, and this allows you to read it as a sentence like it was intended.

But that's just my opinion, changing it to preserve the newlines would be trivial: just add an extra \n when it joins the lines.

skovhus:
> Solution: update to less than 6? Or otherwise it probably doesn't matter much because this test fails based on a user's path

Let us do less than 8. Would be nice that it runs on your machine.

server/src/analyser.ts
```diff
@@ -399,6 +399,67 @@ export default class Analyzer {
...
+
+    // start from the line above
+    let commentBlockIndex = line - 1
+
+    // check if that line starts with a comment
+    // its possible that some lines
+    // might have whitespace before the comment
+    // begins, so check by iterating from
+    // the start of the line
+    const startsWithAComment = (line: string): boolean => {
```
skovhus, line 418: It would be more natural to use a regular expression here.

```
const commentRegExp = /^\s*\#\s*(.*)/g

function getComment(line: string): null | string {
  const matches = commentRegExp.exec(' # this is a comment . ')
  return matches ? matches[1].trim() : null
}
```

Then you can run the while loop until it returns `null`. This also does the newline removal and trimming.

server/src/analyser.ts
```diff
@@ -399,6 +399,67 @@ export default class Analyzer {
...
+      currentLine = doc.getText({
+        start: { line: commentBlockIndex, character: 0 },
+        end: { line: commentBlockIndex + 1, character: 0 },
+      })
+    }
+
+    // since we searched from bottom up, we then reverse
+    // the lines so that it reads top down.
+    const commentStr = commentBlock.reverse().join('\n')
+    return commentStr
```
skovhus, line 460: We should return `null` if there a no comments found.

server/src/server.ts
```diff
@@ -194,9 +194,16 @@ export default class BashServer {
             ? `${symbolKindToDescription(symbol.kind)} defined in ${path.relative(
                 currentUri,
                 symbol.location.uri,
+              )}\n\n${this.analyzer.commentsAbove(
```
skovhus, line 197: We should add `\n\n` if no comment is found (i.e. it returns null).

testing/fixtures/comment-doc-on-hover.sh
```diff
@@ -0,0 +1,39 @@
...
+# has two lines
+func_two() {
+    echo "func_two"
+}
+
+
+# this is not included
+
+# doc for func_three
+func_three() {
```
skovhus, line 37: Let us add another example that doesn't have any immediate documentation. And we could actually make that a variable instead to show that the function also works for variables (not only functions).

server/src/server.ts
nikita-skobov: Do you mean that `commentsAbove` should return `\n\n` instead of null if nothing is found?
Or do you mean that we should not add the `\n\n` between the `defined in X` part and the `commentsAbove` part?

server/src/__tests__/analyzer.test.ts
```diff
@@ -192,6 +192,21 @@ describe('findSymbolCompletions', () => {
   })
 })
 
+describe('commentsAbove', () => {
+  it('returns a string of a comment block above a line', () => {
+    analyzer.analyze(CURRENT_URI, FIXTURES.COMMENT_DOC)
+    expect(analyzer.commentsAbove(CURRENT_URI, 22)).toEqual(' doc for func_one')
```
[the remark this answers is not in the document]
nikita-skobov: remove from the test, or from the actual implementation? ie: when a user has:

```
# doc for func_one
```

the trailing white space is the white space between the `#` and the first word `doc`.

So should all trailing white space be removed within the `commentsAbove` function for each line that has a comment?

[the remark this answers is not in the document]
skovhus: I'm thinking that as a user, I don't need the whitespace in the output, so `commentsAbove` should not return them. The regular expression I posted takes care of this.

server/src/server.ts
skovhus: We should not add `\n\n` if there is no comment. I missed a negation in my first comment.

nikita-skobov: sounds good. What do you think of something like this:

```
`${symbolKindToDescription(symbol.kind)} defined in ${path.relative(
    currentUri,
    symbol.location.uri,
)}${this.analyzer.commentsAbove(
    symbol.location.uri,
    symbol.location.range.start.line,
)}`
```

and if `commentsAbove` does not find any comments above, then it can return an empty string. So that way the tooltip simply displays: `X defined in Y` (and similarly for the section below where `X defined on line Y`)

skovhus: If it always returns a string then you would need to return the new lines from the function. As we want two new lines in case a comment is found. 

Instead the formatting should be done at display time, formatting is not a concern of the commentsAbove function. So I recommend just returning a plain string or a null. 

In case of a null then don’t display, else add two new lines before the string.

nikita-skobov: Ok. I just wanted to minimize the logic from within the template string because it already is quite messy. How about this:

```
// a helper function that either returns the comment string with two new lines
// or an empty string if not found
const getCommentsAbove = (symbol) => {
    const commentString = this.analyzer.commentsAbove(
        symbol.location.uri,
        symbol.location.range.start.line,
    )
    if (commentString) {
        return `\n\n${commentString}`
    }

    // return empty string if commentString is null
    return ''
}


`${symbolKindToDescription(symbol.kind)} defined in ${path.relative(
    currentUri,
    symbol.location.uri,
)}${getCommentsAbove(symbol)}`
```

skovhus: It totally makes sense to extract to a variable and not do it in the template string. 

A string would be fine instead of a function, and the code comment shouldn’t be needed - the code reads nicely. 

```JavaScript

const comment = this.analyzer.commentsAbove(
        symbol.location.uri,
        symbol.location.range.start.line,
 )
const commentString = comment ? `\n\n${comment}` : ''
```

nikita-skobov: :+1: 

sounds good

nikita-skobov: actually looking at it more closely, I think the best way is to use a helper function because the template string is located in a "one-line" filter and map expression, and I want to minimize the amount of code changed in that function

nikita-skobov:
As of commit 1e35af1 I think Ive resolved all issues mentioned in the comments. Please let me know if this can be merged into master, or if there are any other issues.

PR #235, opened 2020-05-15
Use comments as symbol documentation

skovhus:
Iteration based on https://github.com/bash-lsp/bash-language-server/pull/234 (comments above symbols as documentation for onHover) to also use this feature for onCompletion items.


skovhus:
@nikita-skobov I used your work to also return the new symbol documentation when using the onCompletion handler (use for suggestions when writing a symbol matching other symbols in the workspace).

Issue #236, opened 2020-05-16
code completion of variables in quotes


nikita-skobov:
consider the scenario where you have a variable defined, and then you want to use that variable from within quotation marks:

```sh
my_var="world"
echo "hello $my_var"
```

Is it possible to do code completion when you type part of that variable name, consider:
If you type `echo "hello $my_"`, can you get code completion to show up for variables that match `my_`

Is this something that the `vscode-client` extension can potentially do, or is this an issue with vscode itself?

skovhus:
Thanks for reporting this issue. This is a feature we currently do not support, but we probably should. 

The trigger characters needs to be adjusted in the capabilities object. Besides our parser needs to understand the `$` symbol. Currently the `wordAtPoint` returns `null`... we need to figure out if the grammar should be updated or we can work around it.

nikita-skobov:
I will take a look later today/tomorrow, and see if I can try adding this feature. Could you point me to which files/test cases are most relevant to this issue?

skovhus:
@nikita-skobov thanks for offering your help! :) 

I actually already started having a look at this and possibly found some flaws in the `wordAtPoint` logic done in https://github.com/bash-lsp/bash-language-server/pull/192

I'll get back to you. 

But if you want to help out improving the language server and you are curious to learn how the bash parser works, then https://github.com/tree-sitter/tree-sitter-bash needs some attention. This is the parser we use, and as you see it has a bunch of parsing issues currently. Half of our issues is related to the parser. See https://github.com/bash-lsp/bash-language-server/issues?q=is%3Aopen+is%3Aissue+label%3A%22grammar+bug%22

I can give you some pointers if you want to help out:
- https://tree-sitter.github.io/tree-sitter/
- browse some PRs here https://github.com/tree-sitter/tree-sitter-bash and see how the things are being fixed
- read through https://github.com/tree-sitter/tree-sitter-bash/blob/master/grammar.js and get and overview of how the current feature is implemented

skovhus:
Maybe you can try out https://github.com/bash-lsp/bash-language-server/pull/237

I'll test some edges cases and document the PR with some before and after screenshots.

nikita-skobov:
Awesome thanks, I'll give it a go.

skovhus:
This was released as vscode extension 1.10.1 and bash-language-server 1.16.0.

PR #237, opened 2020-05-19
Code completion improvements

skovhus:
Fixes https://github.com/bash-lsp/bash-language-server/issues/236

This PR enables code completion improvements:
- completion support for parameter expansions
- lots of words are now parsed correctly like `else` (we had a bug in the handling around tree-sitter)
- reduces some unnecessary complexities

### Screenshots (before and after)

Screenshot 2020-05-20 10 42 28Screenshot 2020-05-20 10 42 34

Screenshot 2020-05-20 10 43 07Screenshot 2020-05-20 10 42 42

### DIFF
--- a/server/src/__tests__/server.test.ts
+++ b/server/src/__tests__/server.test.ts
@@ -418,4 +418,33 @@ Helper function to add a user",
       ]
     `)
   })
+
+  it('responds to onCompletion with all variables when starting to expand parameters', async () => {
+    const { connection, server } = await initializeServer()
+    server.register(connection)
+
+    const onCompletion = connection.onCompletion.mock.calls[0][0]
+
+    const result: any = await onCompletion(
+      {
+        textDocument: {
+          uri: FIXTURE_URI.SOURCING,
+        },
+        position: {
+          // $
+          line: 14,
+          character: 7,
+        },
+      },
+      {} as any,
+      {} as any,
+    )
+
+    const itemKinds = result.map((item: any) => item.kind)
+
+    // they are all variables
+    expect(itemKinds.every((kind: any) => kind === lsp.CompletionItemKind.Variable)).toBe(
+      true,
+    )
+  })
 })
--- a/server/src/__tests__/server.test.ts
+++ b/server/src/__tests__/server.test.ts
@@ -181,7 +181,7 @@ describe('server', () => {
       {} as any,
     )
 
-    expect(result2).toMatchInlineSnapshot(`null`)
+    expect(result2).toMatchInlineSnapshot(`Array []`)
   })
 
   it('responds to onWorkspaceSymbol', async () => {
@@ -279,9 +279,9 @@ describe('server', () => {
           uri: FIXTURE_URI.INSTALL,
         },
         position: {
-          // else
-          line: 24,
-          character: 5,
+          // empty space
+          line: 26,
+          character: 0,
         },
       },
       {} as any,
@@ -289,7 +289,7 @@ describe('server', () => {
     )
 
     // Entire list
-    expect(result && 'length' in result && result.length > 50).toBe(true)
+    expect(result && 'length' in result && result.length).toBeGreaterThanOrEqual(50)
   })
 
   it('responds to onCompletion with empty list when word is a comment', async () => {
--- a/server/src/__tests__/server.test.ts
+++ b/server/src/__tests__/server.test.ts
@@ -440,11 +440,9 @@ Helper function to add a user",
       {} as any,
     )
 
-    const itemKinds = result.map((item: any) => item.kind)
-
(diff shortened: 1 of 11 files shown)


server/src/__tests__/server.test.ts
```diff
@@ -418,4 +418,33 @@ Helper function to add a user",
...
+      {} as any,
+      {} as any,
+    )
+
+    const itemKinds = result.map((item: any) => item.kind)
+
+    // they are all variables
+    expect(itemKinds.every((kind: any) => kind === lsp.CompletionItemKind.Variable)).toBe(
+      true,
+    )
```
rcjsuen, line 448: Might want to check the length of the `itemKinds` array or at least check that `itemKinds.length > 0`. Otherwise, the test would pass even if `itemKinds` was an empty array (which would be incorrect).

skovhus: Good point.

server/src/server.ts
```diff
@@ -291,22 +296,44 @@ export default class BashServer {
...
+            symbols: shouldCompleteOnVariables
+              ? this.analyzer.getAllVariableSymbols()
+              : this.analyzer.findSymbolsMatchingWord({
+                  exactMatch: false,
+                  word,
+                }),
             currentUri,
           })
 
+    if (shouldCompleteOnVariables) {
```
nikita-skobov, line 333: is this an optimization to avoid creating the massive `allCompletions` array below?
If so, this if statement only returns when the user type:

```
echo $
echo ${
echo "$
echo "${
```

and if the user types anything after that, the value of `shouldCompleteOnVariables` will be false.
Is this desired behavior, or do we want `shouldCompleteOnVariables` to be true when the user has something like:

```sh
start="some variable"
# user begins to type 'st':
# and hopes to autocomplete 'start'
echo ${st

# shouldCompleteOnVariables is false though
```

server/src/server.ts
```diff
@@ -13,6 +13,8 @@ import { BashCompletionItem, CompletionItemDataType } from './types'
 import { uniqueBasedOnHash } from './util/array'
 import { getShellDocumentation } from './util/sh'
 
+const PARAMETER_EXPANSION_PREFIXES = new Set(['$', '${', '${#'])
+
```
nikita-skobov, line 17: this variable is used to test if:

```js
    const shouldCompleteOnVariables = word
      ? PARAMETER_EXPANSION_PREFIXES.has(word)
      : false
```

This check sets `shouldCompleteOnVariables` to true only for the first two, and only if the user hasnt typed anything else. I've done some tests, and I can't find a case where `${#` gets detected.

nikita-skobov:
The PR looks very nice! I tested it out and it works well for most cases that I've tried.

I made a few comments about the variables that it shows for completion, and how basically, it will show you variables if you type: `$` or `${` but if you start typing anything after that, it will show you all variables, builtins, reserved words, etc.

I think this issue is something with tree sitter like you said in #236 

When you type `${a` then the `const word = this.getWordAtPoint` word becomes `a`, instead of `${a`. but maybe this is desired?

In either case, I think it looks good. :+1:

server/src/server.ts
skovhus: Thanks for reading through this carefully.

> is this an optimization to avoid creating the massive allCompletions array below?

This is actually not an optimization. The reason for the early exit is that we only want to suggest parameters/variables here if the word that we complete on is `$` or `${`.

> and if the user types anything after that, the value of shouldCompleteOnVariables will be false.

From my experience, this is not how language server protocol clients work. They get completion for a word (e.g. `$`) once, meaning they do not continuously call the onCompletion while the user types. This makes our life a bit easier here, as we do not need to adjust while the user types out a word.

skovhus:
> I made a few comments about the variables that it shows for completion, and how basically, it will show you variables if you type: $ or ${ but if you start typing anything after that, it will show you all variables, builtins, reserved words, etc.

I don't think this is the case. As I commented above, the client of an LSP will only call once a word is started (e.g `$`), and not while the user types (e.g `$MY_VARIA`).

But if you go to an existing line with an unfinished word (e.g `$MY_VARIA`) and auto completes, then we do not currently know that we are in a parameter expansion, hence we will autocomplete on not only parameters (`MY_VARIABLE`), but also functions, built-ins, etc. This should be a future extension, where we based on the grammar, would see if the current word inside a parameter expansion.

> When you type ${a then the const word = this.getWordAtPoint word becomes a, instead of ${a. but maybe this is desired?

Yes, form the parser's side, `a` is a word inside the parameter expansion. So this is expected behavior.

skovhus:
@rcjsuen @nikita-skobov thanks for the review. Much appreciated! 👏

Issue #239, opened 2020-05-20
PR #237 introduced a brace completion bug for functions


nikita-skobov:
Prior to #237 if you started typing a function, and hit the first brace: `{` then vscode automatically completed it for you, like so:

![example_before](https://user-images.githubusercontent.com/39128800/82475008-727dbd80-9a91-11ea-990d-f74457f2e201.png)

After #237 when you hit the first brace: `{`, it calls our `onCompletion` function and the word is `{`, it finds matching words to `{` and it finds the { command:

![example_after](https://user-images.githubusercontent.com/39128800/82475097-980ac700-9a91-11ea-9c4b-18b3da79beeb.png)

Typically, after typing a brace, a user will want to hit enter to fill in the function body, whereas if the user hits enter, the completion fills in, and the user is left with:

```sh
my_function {{}
```

I have tested this bug out on a couple cases, and it only seems to come up on functions, ie: something like this:

```sh
for i in {}
```

does not trigger the `{` suggested completion,

and 

```sh
echo ${}
```

works as expected.

This error seems to be present with all function declaration syntaxes such as:

```sh
my_function () {
my_function(){
my_function ()
{

function my_function() {
function my_function () {
function my_function {
function my_function ()
{
```

I can make a PR for this, but Im not sure what the best way to resolve this issue is:

My two ideas come to:
1. remove the `{` suggested completion. Is there a use case for the language server telling the user what a `{` does?
2. Otherwise, add a check here:
```js
    if (word) {
      // Filter to only return suffixes of the current word
      return allCompletions.filter(item => item.label.startsWith(word))
    }
```
to check if the `word === '{'` then dont give the `{` completion.

Let me know which of these you think is better (or if theres another approach I haven't considered), and I can send a PR later.

skovhus:
Released as bash-language-server@1.16.1 and vs code extension version 1.10.2.

PR #240, opened 2020-05-20
Fix brace expansion bug

nikita-skobov:
as described in #239 

Please let me know if there's a better way than my `validCompletions.filter` approach.

### DIFF
--- a/server/src/__tests__/server.test.ts
+++ b/server/src/__tests__/server.test.ts
@@ -316,6 +316,30 @@ describe('server', () => {
     expect(result).toEqual([])
   })
 
+  it('responds to onCompletion with empty list when word is {', async () => {
+    const { connection, server } = await initializeServer()
+    server.register(connection)
+
+    const onCompletion = connection.onCompletion.mock.calls[0][0]
+
+    const result = await onCompletion(
+      {
+        textDocument: {
+          uri: FIXTURE_URI.ISSUE101,
+        },
+        position: {
+          // the opening brace '{' to 'add_a_user'
+          line: 4,
+          character: 0,
+        },
+      },
+      {} as any,
+      {} as any,
+    )
+
+    expect(result).toEqual([])
+  })
+
   it('responds to onCompletion when word is found in another file', async () => {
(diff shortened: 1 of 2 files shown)


nikita-skobov:
oops. first push had my merges from master into my fork, and also it failed one of the tests, I fixed it. @skovhus ready for review.

server/src/server.ts
```diff
@@ -388,7 +388,13 @@ export default class BashServer {
 
     if (word) {
       // Filter to only return suffixes of the current word
-      return allCompletions.filter(item => item.label.startsWith(word))
+      const validCompletions = allCompletions.filter(item => item.label.startsWith(word))
```
skovhus, line 391: Thanks for trying fixing this and reporting the error. 

Actually, I cannot see why we ever want to return any completions when the word is `{`. 🤔

So I suggest we simply do an early exit after the `if (word && word.startsWith('#')) {` block with:

```typescript
if (word === '{') {
  // We should not complete when it is not prefixed by a $. This case needs to be
  // as { is a completionProvider triggerCharacter.
  return []
}
```

It would also be great to document that we don't complete on `{`. It could be line `4` in the issue101.sh file.

nikita-skobov: :+1: I redid the commit above, and I added a test case. ready for review.

server/src/__tests__/server.test.ts
```diff
@@ -316,6 +316,30 @@ describe('server', () => {
...
+
+    const onCompletion = connection.onCompletion.mock.calls[0][0]
+
+    const result = await onCompletion(
+      {
+        textDocument: {
+          uri: FIXTURE_URI.ISSUE101,
+        },
+        position: {
+          // inside comment
```
skovhus, line 331: I guess this comment is a copy paste. Usually I write what is on the line and character position.

nikita-skobov: whoops. you're right, my b. Fixed it

skovhus:
@nikita-skobov another great PR! Thanks for helping out 🤩

Issue #220, opened 2020-04-16
Scope aware symbol completion


skovhus:
Currently, when auto-completing, the LSP shows a lot of irrelevant symbols related to the current scope (especially local variables from other function).

Task: figure out if we can use the tree-sitter AST to improve the symbol completion. 

Also see https://github.com/bash-lsp/bash-language-server/issues/190#issuecomment-614260716

skovhus:
@mads-hartmann thoughts?

mads-hartmann:
@skovhus Ideally we'd only show completions for the symbols that are in scope (locally defined in a function, loaded from another file through a `source` invocation, etc.). Doing that correctly would require some pretty sophisticated analysis of the source code though.

So perhaps a smaller first step would be to make it user-configurable what we use to populate the list of completions (local file, workspace, binaries on your path, etc.) :)

mads-hartmann:
If you want to solve completion based on scope then I'd start with function scope.

E.g. in the following snippet, `i_am_local` should not be in the completion list outside of the function. 

```sh
#!/usr/bin/env bash

hello() {
    i_am_global="hello"
    local i_am_local="hej"

    echo "in function: $i_am_global"
    echo "in function: $i_am_local"
}

hello
echo "outside function $i_am_global"
echo "outside function $i_am_local"
```

One way you might do this would be:

1. Don't add `local` variables to the symbols list when we process files
2. When asking for completions, traverse the tree upwards and collect any `local` definitions in scope

Happy to chat more about this approach offline if you're interested :)

skovhus:
Update on file-aware symbol completion.

The grammar currently doesn't have support for sourcing as a type (neither `source ./somefile.sh` or `. ./somefile.sh `) similar to how `variable_assignment` is exposed by the grammar. A first step would be to add this to https://github.com/tree-sitter/tree-sitter-bash including an easy way to retrieve the file name.

Then we could use that for building up a tree of files that source each other, which we could use to make the symbol completion file aware.

PR #244, opened 2020-05-25
Sourcing aware symbols (completion + jump to definition)

skovhus:
Related #220 (not solving `local` scope).

This PR supports:
- internally resolve all files that are sourced (i.e. `. my-file.sh` or `source ~/something`). This is done with a regular expression for now until the grammar supports this natively
- only do auto-completion and jump to definition for symbols in files that are sourced (previously we completed for all symbols found in the workspace)
- support jump-to-definition on the file path used in a source command
- disables globally finding symbols (yay no more weird suggestions)
- the new behavior can be disabled by turning on the `includeAllWorkspaceSymbols` configuration option.

## Update November, 2022

This seems like a nice improvement, and the solution from 2020 has been improved to resolve paths based on the workspace folder and the sourcing file. We also dynamically load files that are sourced so the background analysis is not really needed.

### DIFF
--- a/server/src/util/sourcing.ts
+++ b/server/src/util/sourcing.ts
@@ -1,0 +1,108 @@
+import * as path from 'path'
+import * as LSP from 'vscode-languageserver'
+import * as Parser from 'web-tree-sitter'
+
+import { untildify } from './fs'
+
+// Until the grammar supports sourcing, we use this little regular expression
+const SOURCED_FILES_REG_EXP = /^(?:\t|[ ])*(?:source|[.])\s*(\S*)/gm
+
+export function getSourcedUris({
+  fileContent,
+  fileUri,
+}: {
+  fileContent: string
+  fileUri: string
+}): Set<string> {
+  const uris: Set<string> = new Set([])
+  let match: RegExpExecArray | null
+
+  while ((match = SOURCED_FILES_REG_EXP.exec(fileContent)) !== null) {
+    const relativePath = match[1]
+    const sourcedUri = getSourcedUri({ relativePath, uri: fileUri })
+    if (sourcedUri) {
+      uris.add(sourcedUri)
+    }
+  }
+
+  return uris
+}
+
+/**
+ * Investigates if the given position is a path to a sourced file and maps it
+ * to a location. Useful for jump to definition.
+ * @returns an optional location
+ */
+export function getSourcedLocation({
+  tree,
+  position,
+  uri,
+  word,
+}: {
+  tree: Parser.Tree
+  position: { line: number; character: number }
+  uri: string
+  word: string
+}): LSP.Location | null {
+  // NOTE: when a word is a file path to a sourced file, we return a location to
+  // that file.
+  if (tree.rootNode) {
+    const node = tree.rootNode.descendantForPosition({
+      row: position.line,
+      column: position.character,
+    })
+
+    if (!node || node.text.trim() !== word) {
+      throw new Error('Implementation error: word was not found at the given position')
+    }
+
+    const isSourced = node.previousNamedSibling
+      ? ['.', 'source'].includes(node.previousNamedSibling.text.trim())
+      : false
+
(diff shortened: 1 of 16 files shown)


skovhus:
@nikita-skobov let me know if you want to try this out. I would like your feedback.

server/src/util/sourcing.ts
```diff
@@ -0,0 +1,92 @@
...
+  fileUri,
+}: {
+  fileContent: string
+  fileUri: string
+}): Set<string> {
+  const uris: Set<string> = new Set([])
+  let match: RegExpExecArray | null
+
+  while ((match = SOURCED_FILES_REG_EXP.exec(fileContent)) !== null) {
+    const relativePath = match[1]
```
nikita-skobov, line 21: if user has:

```sh
source "./my_file.sh"
```

Then `relativePath` will contain quotes around it, and the sourcedUri that gets added has quotes at the end of the absolute path, and thus will fail to be read later

nikita-skobov:
I'll play around with this more tomorrow, but for now, as an overall impression I have this to say:

**Please note I am biased** because I created and use a [bash transclusion preprocessor](https://github.com/nikita-skobov/bash-source-combine) which uses an import syntax. Now despite using a preprocessor, this extension, prior to this PR, would still show me onHover, and onCompletion for variables/functions defined in other files. **This is/was very convenient to me** because I can write a script like:

```sh
import * from some_file.sh

some_funct
``` 

And as I start typing `some_funct`, it will give me autocompletion suggestions based on all files in the vscode project directory. But after this PR, this would no longer be the case because it's not a source statement, but rather an import.

Also, in other projects of mine, I have source statements that are dynamic based on other variables, consider:

```sh
if [[ -z $SCRIPTPATH ]]; then SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )" ; fi
LIBPATH="$SCRIPTPATH/../../lib/lib.sh"
if [[ -z $__COMPLETION_LIB_LOADED ]]; then source "$LIBPATH" ; fi
```

Even though the lib.sh file is within the same repository, and contains functions that are used by many scripts in this project, this PR would prevent those functions from getting autocompletion/onhover.


**I realize that many people use this project, and I don't want my bias from getting in the way of a good feature being added**

Some ideas to consider before merging this PR:

- maybe get feedback from other users based on their usage of this extension (idk how we'd ask for feedback other than maybe pinning something on the README :woman_shrugging: )
- maybe modify the `contributes.configuration` like you did with
https://github.com/bash-lsp/bash-language-server/blob/c7d48697c048001477c7d61ad72490fe3652f980/vscode-client/package.json#L34
and add an option to disable `sourceAware`
- in addition to an option to disable `sourceAware`, maybe add a `sourcePattern` which will contain keywords like `.`,`source` and users can add other keywords (eg: I would want to add `import`), and then the `SOURCED_FILES_REG_EXP` in `util/sourcing.ts` would need to be created dynamically based on the value read from the users configuration.


Lastly, I just have a question: is this an optimization? Is it for performance, or user experience?

In the gif posted above it seems you are showing an example where you have to scroll quite a bit until you find the variables you are looking for. Idk about others, but my experience with code completion is I usually start typing the variable/function I want to use, and if it doesnt show up right away, I type another character. Usually after 2 or 3 characters there's only one or two left in the list, and I can hit enter.

(PS. How do you make gifs of code? I've done so before and It envolved `screen recording -> ffmpeg -> convert to gif` which is a long process. Is there a good tool that lets you easily take gifs of your code?)

skovhus:
> Lastly, I just have a question: is this an optimization? Is it for performance, or user experience?

My primary concern with this project (and most software I write) is the user experience. :) I want the server to be as useful as possible... and I considered that completing on all symbols in the workspace to be annoying for some users.

My main challenge with this project is that I only use bash for smaller scripts (so the current implementation where we return all workspace symbols works for me). 

You raised a lot of great points here: dynamic imports, usage of preprocessors (tbh I was not aware of these).

PR #276, opened 2021-01-12
Add instruction for neovim built-in lsp

ekickx:
Hi, I want to add instruction for using the language server with neovim's built-in lsp.


skovhus (approved):
Thanks for the contribution!

PR #281, opened 2021-02-13
Fix incorrect hover documentation

hak33m16:
Fixes hover documentation being incorrect for symbols that are already defined on path and adds in a test to verify that's the case.

However, the reason I'm filing as a draft PR is that I realized something interesting... The hover information won't be entirely accurate if the path symbol is used, then redefined and used in reference to the new definition.

For example:
```
ls

ls() {
  echo "ls override"
}

ls
```

The first _and_ last `ls` will say that they're the "function defined on line 3" even though the first `ls` will call the actual path executable.

Wanted to ping you @skovhus and see if you had any ideas for what a good way of handling a sort of context-aware version of this PR? I only immediately see obvious ways to know if something is defined or isn't, since we don't have access to things as they're being parsed. I could be wrong/misunderstanding that aspect of it, though


skovhus (commented):
I'm so sorry for the late feedback here. Awesome that you fixed this. I don't have any immediate suggestion on how to fix the issue that you uncovered.

> The first and last ls will say that they're the "function defined on line 3" even though the first ls will call the actual path executable.

Is the issue the same if the function was not redefining symbols on the path? In case this is already an issue, then I suggest we merge this PR and create a new issue for fixing the other issue.

testing/fixtures.ts
```diff
@@ -24,8 +24,7 @@ export const FIXTURE_URI = {
   PARSE_PROBLEMS: `file://${path.join(FIXTURE_FOLDER, 'parse-problems.sh')}`,
   SOURCING: `file://${path.join(FIXTURE_FOLDER, 'sourcing.sh')}`,
   COMMENT_DOC: `file://${path.join(FIXTURE_FOLDER, 'comment-doc-on-hover.sh')}`,
-  OPTIONS: `file://${path.join(FIXTURE_FOLDER, 'options.sh')}`,
-  SHELLCHECK_SOURCE: `file://${path.join(FIXTURE_FOLDER, 'shellcheck', 'source.sh')}`,
+  OVERRIDE_SYMBOL: `file://${path.join(FIXTURE_FOLDER, 'override-executable-symbol.sh')}`,
```
skovhus, line 27:
```suggestion
  OPTIONS: `file://${path.join(FIXTURE_FOLDER, 'options.sh')}`,
  SHELLCHECK_SOURCE: `file://${path.join(FIXTURE_FOLDER, 'shellcheck', 'source.sh')}`,
  OVERRIDE_SYMBOL: `file://${path.join(FIXTURE_FOLDER, 'override-executable-symbol.sh')}`,
```

PR #294, opened 2021-06-09
Complete command line arguments

### LINKED ISSUE #117: Integrate with bash-completion, if available
It would be helpful if `bash-language-server` could utilize `bash-completion` to fetch the available switches and arguments for certain commands. I'm not familiar with Typescript myself, unfortunately, so I cannot determine the feasibility of implementing such a feature at this time.

## Subcommand and switch completion

For example, if `git` and its associated Bash completions are installed on the machine, then typing the following into a buffer:

```
git <ctrl+space>
```

would yield the following completions from the language server:

```
add
am
archive
bisect
etc...
```

Similarly, typing the following:

```
git --<ctrl+space>
```

would yield the following completions:

```
--bare
--exec-path
--exec-path=
--git-dir=
--help
etc...
```

Optionally, the server could attempt to fetch results for both `git `, `git -`, and `git --` (in that order) and show completions for all three possibilities such that they can be easily filtered and selected at a glance. I'm not sure about the performance implications of this behavior, however.

## Argument completion

A similar process could allow for argument completion, if the completion function provides it. For example, if `openssh` and its associated Bash completions are installed, then typing the following:

```
ssh <ctrl+space>
```

would possibly yield the following completions:

```
127.0.0.1
12.123.123.123
::1
bar
baz.com
foo.local
github.com
localhost
```

## Future work

If a manpage exists on the system for a specific command (the same one manpage currently used for documentation suggestions in `bash-language-server`), we could potentially search the manpage for the subcommands/switches provided to us by `bash-completion`, such that entering the following:

```
git --h<ctrl+space>
```

would allow the user to select `--help` from the list, and hovering over the `--help` would parse the `git` manpage for a section titled `--help` and display the following:

```
--help
           Prints the synopsis and a list of the most commonly used commands. If the option --all or -a is given then all available commands are printed. If a
           Git command is named this option will bring up the manual page for that command.

           Other options are available to control how the manual page is displayed. See git-help(1) for more information, because git --help ...  is converted
           internally into git help ....
```

This could be done using [explainshell](https://github.com/idank/explainshell) for help, as also suggested in #107.

otreblan:
Closes #117.

https://user-images.githubusercontent.com/39320840/121274961-3a5e3e00-c891-11eb-881b-c85bdee29d10.mp4

In vim, `-` must be added to the keyword characters for this to work.
```vim
autocmd FileType sh set iskeyword+=45
```

### DIFF
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -uo pipefail
+
 DATADIR="$(pkg-config --variable=datadir bash-completion)"
 
 source "$DATADIR/bash-completion/bash_completion"
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+source /usr/share/bash-completion/bash_completion
+
+COMP_LINE="$*"
+COMP_WORDS=("$@")
+COMP_CWORD="${#COMP_WORDS[@]}"
+((COMP_CWORD--))
+COMP_POINT="${#COMP_LINE}"
+COMP_WORDBREAKS='"'"'><=;|&(:"
+
+_command_offset 0 2> /dev/null
+
+if (( ${#COMPREPLY[@]} == 0 ))
+then
+	_longopt "${COMP_WORDS[0]}"
+fi
+
+printf "%s\t" "${COMPREPLY[@]}"
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -15,7 +15,12 @@ _command_offset 0 2> /dev/null
 
 if (( ${#COMPREPLY[@]} == 0 ))
 then
-	_longopt "${COMP_WORDS[0]}"
+	# Disabled by default because _longopt executes the program
+	# to get its options.
+	if (( ${BASH_LSP_COMPLETE_LONGOPTS} == 1 ))
+	then
+		_longopt "${COMP_WORDS[0]}"
+	fi
 fi
 
 printf "%s\t" "${COMPREPLY[@]}"
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -2,6 +2,12 @@
 
 DATADIR="$(pkg-config --variable=datadir bash-completion)"
 
+# Exit if bash-completion isn't installed.
+if (( $? != 0 ))
+then
+	exit 1
+fi
+
 source "$DATADIR/bash-completion/bash_completion"
 
 COMP_LINE="$*"
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -u
-
 DATADIR="$(pkg-config --variable=datadir bash-completion)"
 
 source "$DATADIR/bash-completion/bash_completion"
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-source /usr/share/bash-completion/bash_completion
+DATADIR="$(pkg-config --variable=datadir bash-completion)"
+
+source "$DATADIR/bash-completion/bash_completion"
 
 COMP_LINE="$*"
 COMP_WORDS=("$@")
--- a/server/src/get-options.sh
+++ b/server/src/get-options.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -uo pipefail
+set -u
 
 DATADIR="$(pkg-config --variable=datadir bash-completion)"
 
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -1,6 +1,8 @@
 import * as path from 'path'
 import * as TurndownService from 'turndown'
 import * as LSP from 'vscode-languageserver'
+import * as Process from 'child_process'
+import * as Path from 'path'
 import { TextDocument } from 'vscode-languageserver-textdocument'
 
 import Analyzer from './analyser'
@@ -117,6 +119,16 @@ export default class BashServer {
     params: LSP.ReferenceParams | LSP.TextDocumentPositionParams,
   ): string | null {
     return this.analyzer.wordAtPoint(
+      params.textDocument.uri,
+      params.position.line,
+      params.position.character,
+    )
+  }
+
+  private getCommandNameAtPoint(
+    params: LSP.ReferenceParams | LSP.TextDocumentPositionParams,
+  ): string | null {
+    return this.analyzer.commandNameAtPoint(
       params.textDocument.uri,
       params.position.line,
       params.position.character,
@@ -323,6 +335,22 @@ export default class BashServer {
       return []
     }
 
+    let options: string[] = []
+    if (word && word.startsWith('-')) {
+        const commandName = this.getCommandNameAtPoint({
+          ...params,
+          position: {
+            line: params.position.line,
+            // Go one character back to get completion on the current word
+            character: Math.max(params.position.character - 1, 0),
+          },
+        })
+
+        if(commandName) {
+            options = getCommandOptions(commandName, word)
+        }
+    }
+
     const currentUri = params.textDocument.uri
 
     // TODO: an improvement here would be to detect if the current word is
@@ -385,11 +413,21 @@ export default class BashServer {
       },
     }))
 
+    const optionsCompletions = options.map(option => ({
+      label: option,
+      kind: LSP.SymbolKind.Interface,
+      data: {
+        name: option,
+        type: CompletionItemDataType.Symbol
+      },
+    }))
+
     const allCompletions = [
       ...reservedWordsCompletions,
       ...symbolCompletions,
       ...programCompletions,
       ...builtinsCompletions,
+      ...optionsCompletions,
     ]
 
     if (word) {
@@ -531,3 +569,13 @@ const getMarkdownContent = (documentation: string): LSP.MarkupContent => ({
   // Passed as markdown for syntax highlighting
   kind: 'markdown' as const,
 })
+
+function getCommandOptions(name: string, word: string): string[] {
+    // TODO: The options could be cached.
+    const options = Process.spawnSync(
+        Path.join(__dirname, '../src/get-options.sh'),
+        [name, word]
+    )
+
+    return options.stdout.toString().split('\t')
+}
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -1,8 +1,8 @@
+import * as Process from 'child_process'
 import * as path from 'path'
+import * as Path from 'path'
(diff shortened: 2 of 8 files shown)


skovhus:
@otreblan thanks for adding this!  👏 Do you have time to add a few unit tests? That would be great.

otreblan:
> @otreblan thanks for adding this! clap Do you have time to add a few unit tests? That would be great.

Done

server/src/get-options.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
skovhus, line 1: Shouldn't we add `set -euo pipefail`?

server/src/get-options.sh
```diff
@@ -0,0 +1,19 @@
...
+
+source /usr/share/bash-completion/bash_completion
```
skovhus, line 3: This is not found on OS X. Wondering if the is a more generic way of getting the bash completions. We do have access to the man/help page already. 🤔

server/src/get-options.sh
```diff
@@ -0,0 +1,19 @@
...
+COMP_CWORD="${#COMP_WORDS[@]}"
+((COMP_CWORD--))
+COMP_POINT="${#COMP_LINE}"
+COMP_WORDBREAKS='"'"'><=;|&(:"
+
+_command_offset 0 2> /dev/null
+
+if (( ${#COMPREPLY[@]} == 0 ))
+then
+	_longopt "${COMP_WORDS[0]}"
```
skovhus, line 16: `_longopt` is not available on OS X.

server/src/server.ts
```diff
@@ -531,3 +569,13 @@ const getMarkdownContent = (documentation: string): LSP.MarkupContent => ({
...
   kind: 'markdown' as const,
 })
+
+function getCommandOptions(name: string, word: string): string[] {
+  // TODO: The options could be cached.
+  const options = Process.spawnSync(Path.join(__dirname, '../src/get-options.sh'), [
+    name,
+    word,
+  ])
+
```
skovhus, line 579: We should handle the process not returning a successful result.

server/src/server.ts
```diff
@@ -531,3 +569,13 @@ const getMarkdownContent = (documentation: string): LSP.MarkupContent => ({
...
+  return options.stdout.toString().split('\t')
```
skovhus, line 580: It seems we should trim and filter the empty results here:
```suggestion
  return options.stdout
    .toString()
    .split('\t')
    .map(l => l.trim())
    .filter(l => l.length > 0)
```

skovhus (commented):
Thank you so much for contributing here. This would be a great addition to the language server.

The `get-options.sh` shell script doesn't seem that portable. I'm wondering if we can come up with a more portable/cross platform solution. Note the we already have the man and help page that we could parse. 

Another thing: updating the `triggerCharacters` configuration option (line 107 in `src/server.ts`) to include `-` makes it work without additional configuration.

server/src/get-options.sh
otreblan: The completion doesn't work with `set -e`.
https://man.archlinux.org/man/set.1p#APPLICATION_USAGE

server/src/get-options.sh
otreblan: Now `_longopt` is only executed when `$BASH_LSP_COMPLETE_LONGOPTS == 1`

otreblan:
> Another thing: updating the `triggerCharacters` configuration option (line 107 in `src/server.ts`) to include `-` makes it work without additional configuration.

When `-` is in `triggerCharacters`:

https://user-images.githubusercontent.com/39320840/121825457-18413300-cc78-11eb-8f53-3251f77be46f.mp4

server/src/get-options.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-source /usr/share/bash-completion/bash_completion
+DATADIR="$(pkg-config --variable=datadir bash-completion)"
```
skovhus, line 3: How does the script handle if the user doesn't have the completion installed?

```
~ pkg-config --variable=datadir bash-completion

package bash-completion was not found in the pkg-config search path.
Perhaps you should add the directory containing `bash-completion.pc'
to the PKG_CONFIG_PATH environment variable
No package 'bash-completion' found

~ echo $?
1
```

skovhus: At least on OS X the suggested solution doesn't work. Not sure if the solution works on most unix distros?

otreblan: It should work if the `bash-completion` package is installed and it provides a `bash-completion.pc` file.

server/src/get-options.sh
```diff
@@ -0,0 +1,32 @@
...
+_command_offset 0 2> /dev/null
+
+if (( ${#COMPREPLY[@]} == 0 ))
+then
+	# Disabled by default because _longopt executes the program
+	# to get its options.
+	if (( ${BASH_LSP_COMPLETE_LONGOPTS} == 1 ))
+	then
+		_longopt "${COMP_WORDS[0]}"
+	fi
```
otreblan, line 29: Btw `_longopts` is disabled until `BASH_LSP_COMPLETE_LONGOPTS=1`, because it can execute an arbirary program.

skovhus: Ah, so we require people to set `BASH_LSP_COMPLETE_LONGOPTS=1` in their bash profile, or?

gegoune:
Should it work when editing sh file (with `#!/usr/bin/env bash`) in neovim started from zsh? I have installed bash-completion package with homebrew, but it does not provide completions on `find -`. bashls updated to 2.0.0.

otreblan:
> Should it work when editing sh file (with `#!/usr/bin/env bash`) in neovim started from zsh? I have installed bash-completion package with homebrew, but it does not provide completions on `find -`. bashls updated to 2.0.0 and `iskeyword` contains `-` (45).

Do you have `bash-completion@1` or `bash-completion@2`?

gegoune:
Version 1. Works with @2, thank you! Although I am only getting completions for find, none for head or sort, as per video above. Using native nvim LSP client.

otreblan:
> Version 1. Works with @2, thank you! Although I am only getting completions for find, none for head or sort, as per video above. Using native nvim LSP client.

image

gegoune:
I see, there is no completions for those in `/usr/local/share/bash-completion/completions/`. Commands with completions there do work fine. Great job, thanks!

David-Else:
Hi, I can't find any documentation on this feature in the README.

I have the following installed, and it works with `find - [tab tab]` on the command line:

```
Installed Packages
Name         : bash-completion
Epoch        : 1
Version      : 2.7
Release      : 5.el8
Architecture : noarch
Size         : 895 k
Source       : bash-completion-2.7-5.el8.src.rpm
Repository   : @System
From repo    : anaconda
Summary      : Programmable completion for Bash
URL          : https://github.com/scop/bash-completion
License      : GPLv2+
Description  : bash-completion is a collection of shell functions that take advantage
             : of the programmable completion feature of bash.
```
It does not seem to work in Neovim using the default config for the LSP https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#bashls and https://github.com/hrsh7th/nvim-cmp . Can anyone help? Thanks.


```
npm ls -g --depth=0
/usr/local/lib
├── bash-language-server@2.0.0
```

otreblan:
@David-Else Whats the output of `pkg-config --variable=datadir bash-completion`?

David-Else:
@otreblan The result is nothing, just an empty line.

otreblan:
@David-Else Does your installation has `/usr/share/pkgconfig/bash-completion.pc`?

David-Else:
@otreblan yes:

```
$ls /usr/share/pkgconfig/
total 48K
drwxr-xr-x.   2 root root 4.0K Aug 16 11:42 .
drwxr-xr-x. 233 root root  12K Jul 21 09:17 ..
(various files)
-rw-r--r--.   1 root root  292 Apr 12 01:35 bash-completion.pc
```

otreblan:
@David-Else And it's contents?

David-Else:
@otreblan 

```
bat /usr/share/pkgconfig/bash-completion.pc 
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /usr/share/pkgconfig/bash-completion.pc
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ prefix=/usr
   2   │ compatdir=/etc/bash_completion.d
   3   │ completionsdir=${prefix}/share/bash-completion/completions
   4   │ helpersdir=${prefix}/share/bash-completion/helpers
   5   │ 
   6   │ Name: bash-completion
   7   │ Description: programmable completion for the bash shell
   8   │ URL: https://github.com/scop/bash-completion
   9   │ Version: 2.7
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
Am I still meant to add `autocmd FileType sh set iskeyword+=45`, I assumed that was fixed somewhere in this PR?

otreblan:
@David-Else Your bash-completion is too old, datadir was added 2 years ago https://github.com/scop/bash-completion/blame/master/bash-completion.pc.in

David-Else:
@otreblan Cheers! I will try to update using the source from Github, I am using Rocky Linux 8, and they do have some out of date things hidden in there.

David-Else:
@otreblan Hello again. I removed my old Centos 8 RPM version of `bash-completion` and installed the latest from source. Now my `bash-completion.pc` is installed in `/usr/local/share/pkgconfig` and says:
```
prefix=/usr/local
datadir=/usr/local/share
sysconfdir=/usr/local/etc

compatdir=${sysconfdir}/bash_completion.d
completionsdir=${datadir}/bash-completion/completions
helpersdir=${datadir}/bash-completion/helpers

Name: bash-completion
Description: programmable completion for the bash shell
URL: https://github.com/scop/bash-completion
Version: 2.11
```
I am still not getting auto completion in neovim. You asked me before did I have `/usr/share/pkgconfig/bash-completion.pc`, now the file has moved could that be the problem? Thanks.

David-Else:
@otreblan Thanks to the new 2.1 version it now works! :)

David-Else:
This feature works in Helix, but there is a problem. It does not happen in Neovim, can it be solved here or should I make a request to the Helix devs? If so, what should I ask them to implement? Cheers!

1. Type `find -` and `ctrl-x` to trigger autocomplete:
Screenshot from 2023-01-04 17-27-58

2. Select `depth`

Screenshot from 2023-01-04 17-28-01

3. Now you are left with the original `-` that should not be there, it should be `-depth` not `--depth`

Screenshot from 2023-01-04 17-28-03

skovhus:
@David-Else it seems like a Helix issue as it works in all other clients. And I'm honestly not sure what the issue is.

David-Else:
@skovhus Thanks for the quick reply!

PR #308, opened 2021-08-11
Upgrade tree-sitter-bash to 0.19.4 (break node 10 compatibility)

skovhus:
It seems to only be a "behind the scenes" upgrade.

Turns out the changes here might deprecate node 10, as CI fails with "TypeError: WebAssembly.Global is not a constructor"


rcjsuen:
You may want to skip 11.x as it became [end-of-life on 2019-06-01](https://github.com/nodejs/Release/blob/7d7eef4b2f4922a00f59b3481bd732970ea6bd42/README.md). I would suggest running builds on 12.x, 14.x, and 16.x instead.

skovhus:
> You may want to skip 11.x as it became end-of-life on 2019-06-01. I would suggest running builds on 12.x, 14.x, and 16.x instead.

Thanks! That is a good suggestion.

server/package.json
```diff
@@ -15,7 +15,7 @@
     "url": "https://github.com/bash-lsp/bash-language-server"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=11.0.0"
```
rcjsuen, line 18: @skovhus You should probably bump this up to `>=12.0.0`.

skovhus: Thanks for the careful review. But technically we do support node 11.

Issue #347, opened 2022-04-13
onHover documentation format


lnxbil:
Hi everyone,

Thank you for writing `bls`. I wanted to discuss what possible formats can/should be allowed for functions.

After reading closed issues, I discovered, that the comment can be markdown, as long as it does not have newlines (unfortunately also empty comment lines and not only empty lines) it is interpreted as one comment for hover block.

In general, such comment can be created and it is displayed correctly (at least in the rendered form):

![grafik](https://user-images.githubusercontent.com/727711/163203689-061fa912-bed8-4f22-abf1-9248185581fe.png)

The `---` is there to divide the location of the code from the actual comment and looks nicer to my eyes.

In contrast to that, the built-in hover entries for default bash stuff like `declare` looks like this:

![grafik](https://user-images.githubusercontent.com/727711/163200940-1640a297-9bc1-4f1f-b452-7619e172a885.png)

I was not able to create a comment for my own functions that looks like the built-in. AFAIK, the built-in text is directly taken from either the manpage or the info page, so it is already kind of preformatted and good to go.

Unfortunately, there is no `@javadoc` for Bash that I know of. There are some projects around that define some form of commenting system, but nothing as good as the manpages of the built-ins.

Is anyone aware of a nice and well-established commenting system for bash? What about integrating this into `bls`?
What are you using for commenting system and how well does that integrate itself?

If the arguments could be extracted from such a "better comment", it would also be very nice to have argument completion or just some help that shows what the current argument under the cursor if for.

lnxbil:
First time using TypeScript and it solves my problem. There may exist a better / cleaner / shorter solution, but this fixes my problem.

lnxbil:
In the end, it looks like this:

![grafik](https://user-images.githubusercontent.com/727711/163408491-452d04e8-8a5d-4a89-97c7-60f2d8a95afc.png)

lnxbil:
and it is merged. Thank you.

PR #348, opened 2022-04-14
Adding Multiline Comments

lnxbil:
The overlay should be able to display comments that belong together,
even when an empty comment line is in between, e.g.

    # foo
    #
    # bar

should be allowed and this patch enables this.

(may solve #347)


skovhus (approved):
Thanks for the contribution! I simplified the code a bit by checking if the return value was null.

lnxbil:
> Thanks for the contribution! I simplified the code a bit by checking if the return value was null.

Yes, perfekt. Thank you.

PR #352, opened 2022-04-26
Add Helix to clients

David-Else:
I did an alphabetical sort of the clients to avoid any preferential treatment in the order, and added Helix that supports bash-language-server out of the box. You just need to have it installed and it picks it up: https://docs.helix-editor.com/lang-support.html


skovhus (approved):
Thanks!

PR #396, opened 2022-05-14
Fix analyzer not being called when getHighlightParsingError is off

shabbyrobe:
My apologies, looks like I let a regression slip through. I inadvertently allowed the call to the analyzer to be gated behind the `getHighlightParsingError` function, so if it was off, document trees were never being analysed and cached, breaking lots of things.

It's a bit hard to write a test for at the moment, given `config` is global. @skovhus, are you comfortable with me replacing accesses to the global config functions with an interface type passed in as a dependency to `BashServer`? This will make it trivial to inject different values in a test case. Is there a better way to do this?

There was also an issue where TypeScript seems to not notice that `undefined` can come out of `uriToTreeSitterTrees`, and we were trying to access `rootNodes` against a document even if it wasn't found.


skovhus (approved):
Thanks for the quick fix! 👏 It would be great to refactor this to enable testing. I have no strong opinions on the best option here.

skovhus:
What I suggest is that we reconsider if we should just get rid of the config option. This really is a workaround as the tree sitter grammar isn’t complete. Any downsides to removing it? A user will not get feedback on why some LSP features doesn’t work. 🤔

To test the behaviour we should be able to mock the analyzer and lint function.

shabbyrobe:
> What I suggest is that we reconsider if we should just get rid of the config option.

I have been keeping it turned off actually, it wasn't typically telling me anything shellcheck wasn't doing a better job of telling me. Maybe I need to spend a bit more time with it turned on to see if I really feel this way.

shabbyrobe:
> Thanks for the quick fix! 👏 It would be great to refactor this to enable testing. I have no strong opinions on the best option here.

No worries! My own fault, glad I ran into it before it inconvenienced lots of folks. As to the config thing, I've got an idea, is it OK if I submit a draft PR?

skovhus:
> I've got an idea, is it OK if I submit a draft PR?

anytime!

PR #421, opened 2022-06-01
Add install instructions for Fedora

RMPR:


skovhus:
Thanks! I realize now that we should also update the Mac OS instruction to suggest using `brew install bash-language-server`.

https://formulae.brew.sh/formula/bash-language-server

PR #433, opened 2022-06-17
Fix file uri under Windows are incorrect

jfcherng:
https://en.wikipedia.org/wiki/File_URI_scheme#Windows_2

You can't just prefix it with "file://".
Under Windows, that means it's a network location.

Let's just uses Node's API to do a proper conversion.

https://nodejs.org/api/url.html#urlpathtofileurlpath

Signed-off-by: Jack Cherng <jfcherng@gmail.com>

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bash Language Server
 
+## 3.0.4
+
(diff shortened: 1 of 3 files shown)


skovhus (approved):

skovhus:
Thanks for fixing this!

PR #443, opened 2022-06-28
Fix no diagnostics without WorkspaceFolders support

a5ob7r:
Document diagnostics by shellcheck doesn't work without WorkspaceFolders support clients, because no response from the client about WorkspaceFolders and it blocks subsequent steps. To resolve this, send the request only if the client supports WorkspaceFolders.


server/src/server.ts
```diff
@@ -53,17 +53,20 @@ export default class BashServer {
 
   private documents: LSP.TextDocuments<TextDocument> = new LSP.TextDocuments(TextDocument)
   private connection: LSP.Connection
+  private clientCaps: LSP.ClientCapabilities
```
skovhus, line 56: Nitpick: but let us spell this out. clientCapabilities.

skovhus (commented):
Thanks for fixing this. 

I’m curious: Which clients doesn’t support this?

server/src/server.ts
a5ob7r: I got it.

a5ob7r:
> I’m curious: Which clients doesn’t support this?

It's [vim-lsp](https://github.com/prabirshrestha/vim-lsp). Actually it supports WorkspaceFolders, but that feature is experimental and disabled by default now.
https://github.com/prabirshrestha/vim-lsp/blob/da80b018311b811eceac86fd0a59c521ffa9d2ab/doc/vim-lsp.txt#L2033

skovhus (approved):

skovhus:
Released as 3.0.5. Does it work for you @a5ob7r?

a5ob7r:
@skovhus Yes, it works fine. Thank you!

PR #466, opened 2022-07-23
Fix completion kind info

### LINKED ISSUE #343: KindInfo in auto-completion seems to be messed up
When typing `if` this language server returns `[{'label': 'if', 'data': {'name': 'if', 'type': 2}, 'kind': 11}]`

The `11` means `CompletionItemKind.Unit`, but it should return `14` `CompletionItemKind.Keyword`.

grafik

A look at server.ts reveals it being typed as `SymbolKindInfo.Interface`, which is valued `11` as well.

https://github.com/bash-lsp/bash-language-server/blob/6ee982669df77a62e2f8588c3634aa28a6226ce2/server/src/server.ts#L386

I have a feeling to see 2 errors here:

1. Instead of `SymbolKindInfo` I think `CompletionItemKind` should be used in all those lines around L386.
2. The kind of reserved words should probably be `CompletionItemKind.Keyword` instead of `SymbolKindInfo.Interface`.

I found a `symbolKindToCompletionKind()` but it is missing a translation for `SymbolKindInfo.Keyword` to `CompletionItemKind.Keyword` thus returning `CompletionItemKind.Text` in such cases. But it is not even applied for completions.

deathaxe:
This PR assigns `LSP.CompletionItemKind` constants to completions.

Fixes #343

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -403,7 +403,7 @@ export default class BashServer {
 
     const reservedWordsCompletions = ReservedWords.LIST.map((reservedWord) => ({
       label: reservedWord,
-      kind: LSP.SymbolKind.Interface, // ??
+      kind: LSP.CompletionItemKind.Keyword,
       data: {
         name: reservedWord,
         type: CompletionItemDataType.ReservedWord,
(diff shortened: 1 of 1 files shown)


skovhus (approved):
Thanks for fixing this!

PR #535, opened 2022-10-29
fix: should check type of folders

### LINKED ISSUE #516: TypeError: folders is not iterable
```log
/usr/local/lib/node_modules/bash-language-server/out/linter.js:58
            for (const folder of folders) {
                                 ^

TypeError: folders is not iterable
    at Linter.<anonymous> (/usr/local/lib/node_modules/bash-language-server/out/linter.js:58:34)
    at Generator.next (<anonymous>)
    at /usr/local/lib/node_modules/bash-language-server/out/linter.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/usr/local/lib/node_modules/bash-language-server/out/linter.js:4:12)
    at Linter.runShellcheck (/usr/local/lib/node_modules/bash-language-server/out/linter.js:56:16)
    at Linter.<anonymous> (/usr/local/lib/node_modules/bash-language-server/out/linter.js:31:39)
    at Generator.next (<anonymous>)
    at /usr/local/lib/node_modules/bash-language-server/out/linter.js:8:71
    at new Promise (<anonymous>)
```

bash-language-server --version
Version is 3.1.0

Shane-XB-Qian:
closes #516

### DIFF
--- a/server/src/linter.ts
+++ b/server/src/linter.ts
@@ -68,8 +68,10 @@ export default class Linter {
       '--external-sources',
       `--source-path=${this.cwd}`,
     ]
-    for (const folder of folders) {
-      args.push(`--source-path=${folder.name}`)
+    if (Array.isArray(folders)) {
+      for (const folder of folders) {
+        args.push(`--source-path=${folder.name}`)
+      }
     }
 
     let out = ''


server/src/linter.ts
```diff
@@ -63,8 +63,10 @@ export default class Linter {
     folders: LSP.WorkspaceFolder[],
   ): Promise<ShellcheckResult> {
     const args = ['--format=json1', '--external-sources', `--source-path=${this.cwd}`]
-    for (const folder of folders) {
-      args.push(`--source-path=${folder.name}`)
+    if (Array.isArray(folders)) {
```
skovhus, line 66: In which case is `folders` not an array? What is it?

I'm asking as the type signature here explicitly says that it should always be an array of `LSP.WorkspaceFolder`

Shane-XB-Qian: according to rough looking to lsp specification, 
looks it said `It can be 'null' if the client supports workspace folders but none are configured`.

skovhus: That case is handled here: https://github.com/bash-lsp/bash-language-server/blob/main/server/src/server.ts#L92

Shane-XB-Qian:
```typescript
      const folders = this.clientCapabilities.workspace?.workspaceFolders
        ? await connection.workspace.getWorkspaceFolders()
        : []
      const lintDiagnostics = await this.linter.lint(change.document, folders || [])
```

1. `this.clientCapabilities.workspace?.workspaceFolders` this got from `initialize` and that is `true`.
2. however `await connection.workspace.getWorkspaceFolders()` looks send `workspace/workspaceFolders` to lsp client, but got `{'id': 0, 'jsonrpc': '2.0', 'result': {}}`
3. so that `folders` looks was a `{}` ?!              <--------- !!!

so the problem is:
1. could it get from ` this.clientCapabilities.workspaceFolders` when/if ` this.clientCapabilities.workspace?.workspaceFolders` is true?
2. should lsp client response with `result: ........` something else or could be `{}`?
3. anyway, `{}` perhaps that's the direct problem.

server/src/linter.ts
Shane-XB-Qian: the problem looks was the previous statement before L92.

Shane-XB-Qian:
and looks others lsp server, i saw they just get workspaces folders from `clientCapabilities.workspaceFolders`, did not send/get `workspace/workspaceFolders`.

Shane-XB-Qian:
@skovhus that should be a bug (or un-impl feature) of lsp client, i've opened a pr to correct/impl that.
// see https://github.com/yegappan/lsp/pull/114

// but this pr to bash lsp server, perhaps should be better merged too, since varied lsp client may send others type of `folders` var.

skovhus:
> but this pr to bash lsp server, perhaps should be better merged too, since varied lsp client may send others type of folders var.

We shouldn't extend the LSP server here to support faulty clients – we want to follow the LSP standard. But thanks for fixing https://github.com/yegappan/lsp/pull/114

I'm closing this for now, but thanks for looking into it.

PR #536, opened 2022-10-29
Add "--shell=bash" argument for ShellCheck

thomasjm:
This PR adds the argument `--shell=bash` when invoking ShellCheck to improve robustness when you don't have a shebang or file extension that ShellCheck can use to determine the shell type. In particular, ShellCheck doesn't automatically assume `.sh` corresponds to Bash, which can lead to getting [SC2148](https://www.shellcheck.net/wiki/SC2148).

Since this language server is Bash-only, I figure it makes sense to always run ShellCheck in this mode.

### DIFF
--- a/server/src/linter.ts
+++ b/server/src/linter.ts
@@ -62,7 +62,7 @@ export default class Linter {
     document: LSP.TextDocument,
     folders: LSP.WorkspaceFolder[],
   ): Promise<ShellcheckResult> {
-    const args = ['--format=json1', '--external-sources', `--source-path=${this.cwd}`]
+    const args = ['--shell=bash', '--format=json1', '--external-sources', `--source-path=${this.cwd}`]
     for (const folder of folders) {
       args.push(`--source-path=${folder.name}`)
     }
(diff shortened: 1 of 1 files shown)


server/src/linter.ts
```diff
@@ -62,7 +62,7 @@ export default class Linter {
     document: LSP.TextDocument,
     folders: LSP.WorkspaceFolder[],
   ): Promise<ShellcheckResult> {
-    const args = ['--format=json1', '--external-sources', `--source-path=${this.cwd}`]
+    const args = ['--shell=bash', '--format=json1', '--external-sources', `--source-path=${this.cwd}`]
```
skovhus, line 65: Can you run `yarn lint` and recommit the file?

thomasjm: Done

skovhus (approved):

adoyle-h:
@skovhus Could you please create a new release? I need this commit while using neovim, but [the package installer only follows the latest release](https://github.com/williamboman/mason.nvim/discussions/672#discussioncomment-4172921). Thanks.

skovhus:
3.1.1 should be out now. Thanks

PR #562, opened 2022-11-26
fix: default shell on Windows should be cmd.exe

jfcherng:
As of `server-3.2.0`, there is auto detection for `shellcheck` executable.

https://github.com/bash-lsp/bash-language-server/blob/fa9120c3b37d19c285fa91d99831231c88b6a874/server/src/linter.ts#L24 

But it uses `bash` by default.

---

But `which` is not available under Windows `cmd.exe`. A proper functionality fix would be a cross-platform solution like [node-which](https://www.npmjs.com/package/which). If we don't want to do this part, I can revert it.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -26,7 +26,8 @@
     "urijs": "^1.19.11",
     "vscode-languageserver": "^6.1.1",
     "vscode-languageserver-textdocument": "^1.0.1",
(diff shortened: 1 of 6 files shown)


jfcherng:
I will re-create a more clean one later.

PR #563, opened 2022-11-26
Fix auto detection for path of `shellcheck.exe` on Windows

jfcherng:
# Issues

- There is no `bash`/`which` on Windows, by default.
- Even if using the `bash`/`which` from git-bash or WSL, the returned path is Unix-style. And Node.js on Windows is not happy with it.

# Solutions

- Use `cmd.exe` as the default shell on Windows.
- Use the `which` package to find `shellcheck` in a cross-platform way.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -26,7 +26,8 @@
     "urijs": "^1.19.11",
     "vscode-languageserver": "^6.1.1",
     "vscode-languageserver-textdocument": "^1.0.1",
-    "web-tree-sitter": "^0.20.5"
+    "web-tree-sitter": "^0.20.5",
+    "which": "^2.0.2"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


skovhus (approved):
Thank you so much for fixing this so fast! This was on my todo list. 👏

skovhus:
@jfcherng any specific reason for using version 2 of the `which` library? I see that there is a version 3, but haven't looked if there are any breaking changes (e.g. like only supporting ESM)

jfcherng:
> @jfcherng any specific reason for using version 2 of the `which` library? I see that there is a version 3, but haven't looked if there are any breaking changes (e.g. like only supporting ESM)

Simply because version 3 requires Node v14, so it will failed tested for Node v12.

PR #585, opened 2022-11-29
Remove error codes from message to avoid dupes

kylo252:
Skip including redundant information in the diagnostic message, since the client can already make
use of the diagnostic code directly.

refs:
  - https://github.com/microsoft/vscode/issues/62159
  - https://github.com/microsoft/vscode/issues/62373


skovhus (approved):
This is great. I’m wondering if we should add the SC prefix to the code to make googling it easier. 

I do have a local version where I’m adding a URL to the diagnostics.

PR #664, opened 2023-01-07
Bash: Add snippets for VS Code

EmilySeville7cfg:
Add snippets for:
- language constructs
- builtins
- expansions
- devices

Motivation: have standardized and easy memorable snippets.


EmilySeville7cfg:
I guess I can do the same job for other supported editors by this extension. :)

skovhus:
Thanks for the contribution!

> I guess I can do the same job for other supported editors by this extension. :)

Yes, if we move this to the server, then all editors will get these snippets available. I believe that is the way to go if we support snippets. 

I’m curious to know how many would prefer having snippets available. I assume we need a way to disable them or potentially override them. As the server is used by a lot of people, we need to think about the reception of this feature. 

Example of other language servers returning snippets: https://github.com/iamcco/vim-language-server/blob/30f442dfa0ff8436eceeb2a306d1e8fd17bbdc6b/src/server/snippets.ts

EmilySeville7cfg:
@skovhus, it's [possible](https://github.com/microsoft/vscode/issues/10565#issuecomment-721832613) to disable them now AFAIK.

EmilySeville7cfg:
Is it possible to move snippets to server a little bit later but now accept these ones? I am not JS/TS developer that's why I am asking for it.

skovhus:
> Is it possible to move snippets to server a little bit later but now accept these ones? I am not JS/TS developer that's why I am asking for it.

That’s a nice and lean approach. Then we might get some feedback before rolling this out to all clients. 

Let me know once you are satisfied with these. I’m not a big snippet developer myself, so I would need to spend some time playing around with these. Maybe there are some standard bash snippets that we can get inspiration from?

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,221 @@
...
+        ]
+    },
+    "alias": {
+        "description": "[alias]",
+        "prefix": "alias",
+        "body": [
+            "alias ${1:name}=${2:value}"
+        ]
+    },
+    "builtin": {
```
skovhus, line 165: We already auto complete on builtins, so I’m unsure what this feature adds. Anything we should do with the current auto completions instead?

EmilySeville7cfg: I didn't know that, I can remove such snippet.

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,214 @@
+{
+    "shebang": {
+        "description": "[sh]hebang",
+        "prefix": "sh",
```
skovhus, line 4: I'm unsure about the convention here, but at least for JavaScript/TypeScript project the snippets usually have a more descriptive name/prefix. I would personally prefer `shebang` compared to `sh`, `while` compared to `wh`, `for` instead of `fo`, etc. 

Screenshot 2023-01-09 at 13 55 29

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,214 @@
...
+        ]
+    },
+    "..": {
+        "description": "brace expansion",
+        "prefix": "..",
+        "body": [
+            "{${1:from}..${2:to}}"
+        ]
+    },
+    "echo": {
```
skovhus, line 137: I'm not sure how much value very basic snippets like `echo` and `source` provides.

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,214 @@
...
+        "prefix": "alias",
+        "body": [
+            "alias ${1:name}=${2:value}"
+        ]
+    },
+    "cd": {
+        "description": "[cd]",
+        "prefix": "cd",
+        "body": [
+            "cd \"${1:path/to/directory}\" || echo \"${2:error_message}\" >&2"
```
skovhus, line 169: This seems rather non-generic.

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,214 @@
...
+        ]
+    },
+    "let": {
+        "description": "[let]",
+        "prefix": "let",
+        "body": [
+            "let ${1:argument}"
+        ]
+    },
+    "test": {
```
skovhus, line 200: `test` seems really useful with all the different options here.

vscode-client/snippets/snippets.json
EmilySeville7cfg: Just to save some keystrokes and quote argument automatically. :)

vscode-client/snippets/snippets.json
EmilySeville7cfg: I can add several possible prefixes for snippets to be able call them quickly and be able find them quickly too.

EmilySeville7cfg: Done. :)

vscode-client/snippets/snippets.json
EmilySeville7cfg: `=~` has been added.

vscode-client/snippets/snippets.json
EmilySeville7cfg: I've also put naming convention in this repo for snippets.

vscode-client/snippets/snippets.json
skovhus: For snippets to be useful I believe they should be fairly generic and applicable for most users. A good example is the `test` snippet that is a very generic. For `cd` we assume that the user wants to handle error case like this and swallow the error (exit code would be 0 after running this). 

I would like us to remove the snippet for `cd` for now.

skovhus (commented):
This is a really good start! Thanks for your work. I've added a few suggestions.

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,249 @@
...
+        "description": "[test]",
+        "prefix": "test",
+        "body": [
+            "[[ ${1:argument1} ${2|-ef,-nt,-ot,==,=,!=,=~,<,>,-eq,-ne,-lt,-le,-gt,-ge|} ${3:argument2} ]]"
+        ]
+    },
+    "device": {
```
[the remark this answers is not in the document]
EmilySeville7cfg: What about to just `dev`?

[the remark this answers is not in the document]
EmilySeville7cfg: I've renamed snippet to `stream`.

EmilySeville7cfg:
> I would like us to remove the snippet for cd for now.

Can I ask what best shell scripting practices to stick with to form expectations of what user might want to type? Can we assume that user don't want to include error handling for `cd`-like commands by default?

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,249 @@
...
+        "body": [
+            "if ${1:command}; then",
+            "\t$0",
+            "fi"
+        ]
+    },
+    "if-else": {
+        "description": "[if] [e]lse",
+        "prefix": [
+            "ife",
```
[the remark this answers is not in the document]
EmilySeville7cfg: > I understand that the intension with these shorthand prefixes like ife is to save keystrokes. I'm personally more a fan of self explanatory prefixes that reduces the cognitive load, compared to saving a few keystrokes.

Can I leave just longer but more readable prefixes like `if-else` instead of smth like `ife`?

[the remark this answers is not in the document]
skovhus: > Can I leave just longer but more readable prefixes like if-else instead of smth like ife?

The long prefixes is what I advocate for here, so let us keep those. It also matches how a language like TypeScript provides exposes snippets in vscode.

vscode-client/snippets/snippets.json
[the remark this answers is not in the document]
skovhus: I'm also not convinced anyone would find them by typing `stream`, so I would just get rid of it. But no strong opinion here. :)

skovhus:
> Can I ask what best shell scripting practices to stick with to form expectations of what user might want to type? Can we assume that user don't want to include error handling for cd-like commands by default?

As there isn't one standard way of doing it and the command is so trivial to type, I don't see that much value in keeping it.

vscode-client/snippets/snippets.json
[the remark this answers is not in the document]
EmilySeville7cfg: Done. :)

vscode-client/snippets/snippets.json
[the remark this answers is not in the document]
skovhus: Ah awesome! 👏

skovhus (approved):

vscode-client/snippets/snippets.json
```diff
@@ -0,0 +1,228 @@
...
+        "description": "[fo]r",
+        "prefix": "for",
+        "body": [
+            "for ${1:variable} in ${2:list}; do",
+            "\t$0",
+            "done"
+        ]
+    },
+    "function": {
+        "description": "[fu]nction",
```
skovhus, line 57: Late comment: Does the `[fu]` naming here in the description make sense? Wasn't that to make the short prefixes easier to understand?

skovhus: And as the names are now rather descriptive, should be also remove the brackets from the ones where the prefix matches the name (e.g `echo` instead of `[echo]`)?

EmilySeville7cfg: It's my mistake, I'll gonna to correct mnemonics soon. :)

EmilySeville7cfg: PR has been created.

PR #667, opened 2023-01-07
Add instructions for integrating with YCM to readme

TriHardStudios:
I personally use YCM and think it would be helpful to include instructions on how to set it up in the readme.


skovhus (approved):
Thanks!

Let me know if you have any feedback or wishes for future features for the bash language server. 🙏

PR #715, opened 2023-02-07
Add fallback for `COMPLETIONSDIR` if `pkg-config` fails

babaric-dev:


server/src/get-options.sh
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
-COMPLETIONSDIR="$(pkg-config --variable=completionsdir bash-completion)"
+COMPLETIONSDIR="$(pkg-config --variable=completionsdir bash-completion || echo "${PREFIX:-/usr}/share/bash-completion/completions")"
```
skovhus, line 3: Is the `PREFIX` used for anything?

skovhus: It would be awesome if we also could this to work for OS X. There bash-completions are usually installed through brew and ends up in `/opt/homebrew/Cellar/bash-completion/1.3_3/etc/bash_completion`, which is sourced through your bashrc file through `/opt/homebrew/etc/profile.d/bash_completion.sh`.

I guess we could either move some of this to node to make it easier to maintain, or add support for testing an array here.

babaric-dev: @skovhus $PREFIX is used when variable is set and default to /usr. A must on systems without /usr as prefix.

babaric-dev: Btw, I don't own a Mac (either im poor or I don't want to get stuck in Apple's ecosystem). So I need someones help for implementing this in MacOS.

babaric-dev:
Superseded by #717

PR #717, opened 2023-02-07
Add fallback mechanism if `pkg-config` fails

babaric-dev:
`COMPLETIONSRC`: path to sourced `bash_completion`

If `pkg-config` fails to get `bash_completion` script:
- use `brew` on macOS
- use `PREFIX` on Linux

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bash Language Server
 
+## 4.7.0
+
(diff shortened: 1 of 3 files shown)


skovhus (approved):
Thanks! I'm surprised by the auto completions though. I usually end up with too many `--` when testing this on Mac OS. But that is another issue. Does it work well for you (which editor)?

PR #763, opened 2023-03-09
server: Only perform dynamic registration if client supports it

the-mikedavis:
The server shouldn't send a `client/registerCapability` unless the client signals support with the dynamicRegistration client capability set to true for didChangeConfiguration. This change checks that field before attempting to register for config change notifications.

This change is important for clients that don't support dynamic capability registration: if we reply to the registerCapability request with a MethodNotHandled JSONRPC error, the promise from `connection.client.register` is rejected which causes the server to exit.

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bash Language Server
 
+## 4.8.3
+
+- Skip sending a `client/registerCapability` request when dynamic capability registration is not supported by the client https://github.com/bash-lsp/bash-language-server/pull/763
+
 ## 4.8.2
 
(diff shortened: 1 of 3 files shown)


skovhus (commented):
Awesome! Thanks for fixing this. 

Are you comfortable bumping the server version inside server/package.json with a minor increase? It would also be great if you can add an entry in the server/CHANGELOG file.

the-mikedavis:
Sure thing! I bumped the patch version since I see this as a small fix but I can bump this to 4.9.0 instead if you prefer

skovhus (approved):
👏

PR #805, opened 2023-03-30
Snippets update: add more snippets and change their naming convention

EmilySeville7cfg:
- document all snippets via `documentation`
- add snippets for commands like `if` with `[[` command
- add more `sed` and `awk` snippets
- change naming convention to make labels more descriptive (but longer)
- add snippet for function doc (closes #781)
- snippets for jq, yq

PR #812, opened 2023-04-05
Bugfix: use "-" instead of "." in snippets

EmilySeville7cfg:
Addresses [this](https://github.com/bash-lsp/bash-language-server/pull/805#issuecomment-1495673783) comment.

@Shane-XB-Qian

### DIFF
--- a/server/src/snippets.ts
+++ b/server/src/snippets.ts
@@ -54,7 +54,7 @@ export const SNIPPETS: BashCompletionItem[] = [
     documentation: 'if operator',
     label: 'if-test',
     insertText: [
-      'if [[ $${1:variable} ${2|-ef,-nt,-ot,==,=~,!=,<,>,-lt,-le,-gt,-ge|} $${3:variable} ]]; then',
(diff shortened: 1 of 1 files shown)


Shane-XB-Qian:
thx, this looks good.

// though lsp client to recognize `[[` `]]` may still had issues, but that maybe not a problem from this.

Shane-XB-Qian:
well, looks i found the reason, though i am not sure which the real one caused it.
but for example `if-else-test`, you set `'if [[ $${1:variable} ${2|-ef,-nt,-ot,==,=~,!=,<,>,-lt,-le,-gt,-ge|} $${3:variable} ]]; then'`,
those two `$$` maybe convenience in some cases, but to shell or to snip plugin, `$` is a special meaning char....

so if you can or make it be simple, 
i wish you could remove that double `$$`, but just one `$`, 
user can manually input the `$` for `var`, Besides that actually Not always a var but a const string/number etc.

EmilySeville7cfg:
Thanks for feedback. Gonna fix this issue today.

Issue #846, opened 2023-05-14
flag suggestions like `--verbose` get triggered by two dashes `--`, but insert four `----`


j-lakeman:
### Code editor

Visual Studio Code

### Platform

Linux

### Version

v1.36.0

### What steps will reproduce the bug?

Insert a command that has the `--verbose` flag, like `rsync`, for example. Continue typing `--v`, until the autocompletion offers `--version` and `--verbose`. Pick either one from the menu. Two additional dashes get inserted, so that there are four altogether.

### How often does it reproduce? Is there a required condition?

reproduces for any `--` flags

### What is the expected behavior?

don't insert additional dashes

### What do you see instead?

Two additional dashes get inserted, so that there are four altogether.
This is unintended behaviour.

### Additional information

[video.webm](https://github.com/bash-lsp/bash-language-server/assets/101660243/91659c2d-aea1-436e-a6c1-912dde72375a)

skovhus:
Thanks for reporting this! 

A fix for this was released in the vscode extension version [1.37.0](https://github.com/bash-lsp/bash-language-server/commit/16412235491855ac1f25014236c35e8d89aecb31).

j-lakeman:
Cheers, seems resolved! But I'm experiencing another issue now. The first letter of the accepted flag gets removed, see video below:

[video.webm](https://github.com/bash-lsp/bash-language-server/assets/101660243/8b80193c-f160-44f0-a7db-be1999e12b7c)

lxr2010:
> Cheers, seems resolved! But I'm experiencing another issue now. The first letter of the accepted flag gets removed, see video below:
> 
>  video.webm

I'm using nvim.cmp and bash-language-server, also have this problem.
It seems that nvim.cmp deals with "insertText" field differently from VS code client. (#847)
I tried to describe insertion with "textEdit" instead of "insertText" , recompiled and reinstalled bash-language-server,  and the problem solved.

PR #861, opened 2023-05-31
Fix option/flag completion via textEdit.

lxr2010:
Update testcases.
Issue #846


skovhus:
Thanks! Released as 4.9.3.

Issue #880, opened 2023-06-24
Crashes in tree-sitter in 4.10.0


rchl:
### Code editor

Sublime Text

### Platform

macOS

### Version

bash-language-server 4.10.0
Node 16.17.1

### What steps will reproduce the bug?

Opened a `.zshrc` file.

The file doesn't seem to have any sensitive information so I'll provide it here but I don't know if it will reproduce without the referenced files.

<details>
<summary>.zshrc</summary>

```sh
#!/bin/zsh

# Enable Powerlevel10k instant prompt. Should stay close to the top of ~/.zshrc.
# Initialization code that may require console input (password prompts, [y/n]
# confirmations, etc.) must go above this block; everything else may go below.
if [[ -r "${XDG_CACHE_HOME:-$HOME/.cache}/p10k-instant-prompt-${(%):-%n}.zsh" ]]; then
  source "${XDG_CACHE_HOME:-$HOME/.cache}/p10k-instant-prompt-${(%):-%n}.zsh"
fi

# Lines configured by zsh-newuser-install
HISTFILE=~/.histfile
HISTSIZE=1000
SAVEHIST=1000
setopt appendhistory
setopt autocd
setopt correct
setopt hist_ignore_all_dups
# setopt print_exit_value
unsetopt beep
bindkey -e
# End of lines configured by zsh-newuser-install
# The following lines were added by compinstall

zstyle ':completion:*' completer _complete _ignored _approximate
zstyle ':completion:*' list-colors ''
zstyle ':completion:*' matcher-list '' 'm:{[:lower:]}={[:upper:]}'
zstyle ':completion:*' max-errors 1
zstyle ':completion:*' menu select
zstyle ':completion:*' select-prompt %SScrolling active: current selection at %p%s
zstyle :compinstall filename ~/.zshrc

fpath=(~/.zsh/zsh-completions/src $fpath)

autoload -Uz compinit
# -u switch to disable insecure dirs warning
compinit -u
# End of lines added by compinstall

# unsetopt BASH_AUTO_LIST
# unsetopt AUTO_MENU

autoload -U select-word-style
select-word-style bash

setopt promptsubst

# Load the prompt theme system
autoload -U promptinit
promptinit

# Use the wunjo prompt theme
# prompt wunjo
# prompt fade green

# place this after nvm initialization!
autoload -U add-zsh-hook
load-nvmrc() {
  local node_version="$(nvm version)"
[...]

Issue #859, opened 2023-05-29
"Go to Definition" rarely works, Bash IDE output is full of "WARNING ⛔️ Error while parsing file"


lkrms:
### Code editor

Visual Studio Code

### Platform

Linux and macOS

### Version

v1.37.0 (last working version was Bash IDE v1.28.0)

### What steps will reproduce the bug?

It's triggered by so many scripts in one of my projects, I can't be sure there aren't other causes, but here's a trivial snippet that triggers the parsing error:

```bash
! VAR=
```

A less trivial example that's closer to how one might actually use a negated exit status:

```bash
! VAR=$(some-command | grep .) || printf '%s\n' "$VAR"
```

Interestingly, statements like `! ((VAR))` are parsed without issue.

### How often does it reproduce? Is there a required condition?

This issue is reproducible 100% of the time.

### What is the expected behavior?

Obviously the code should parse without triggering a syntax error, but in the event that it doesn't, output more specific than "[Warn  - 20:12:15] 10:12:15.379 WARNING ⛔️ Error while parsing file:///path/to/file.sh: syntax error" -- even with the `BASH_IDE_LOG_LEVEL` environment variable set to `debug`, and the VS Code settings below -- would have saved a bunch of time using bisection to identify the problematic syntax 😬

```json
{
  "bashIde.logLevel": "debug",
  "bashIde.enableSourceErrorDiagnostics": true
}
```

### What do you see instead?

- "Go to Definition" and other LSP-serviced features fail for symbols in files that can't be parsed.
- No indication is given that anything is wrong. Problematic files appear to be problem-free in VS Code.
- In VS Code's output panel, "Bash IDE" reports "WARNING ⛔️ Error while parsing file:///path/to/file: syntax error" for many files during background scanning, and they continue whenever I'm editing a problematic file.

skovhus:
The parser used (tree-sitter) is by no means perfect. It keeps improving. But we specifically do no show parsing errors in the UI (as diagnostics) as there are too many false positives. 

For most users the LSP works well for most script, but you might be using some bash constructs that are not supported yet in trees-sitter.

skovhus:
I’m unsure how we can provide a better error here. Any suggestions are more than welcome.

LeonardoMor:
Same in NeoVim. I have a `main` function on a script that calls other functions. Some are in a file that I am sourcing. Jump to definition works on all the functions located in the external file and only works on the main function itself in the local file. It's very strange.

[...]

PR #881, opened 2023-06-26
Handle tree sitter parsing crashes gracefully

skovhus:
Closes https://github.com/bash-lsp/bash-language-server/issues/880


mataha:
I believe this would solve #833 and #859 as well.

skovhus:
It would not solve the missing symbol or improve the parser, but it would remove the unhandled promise rejection from the logs.

PR #897, opened 2023-07-05
Feature/better external command support

EmilySeville7cfg:
Added snippets for:

- zip, unzip
- find

### DIFF
--- a/server/src/snippets.ts
+++ b/server/src/snippets.ts
@@ -376,6 +376,31 @@ export const SNIPPETS: BashCompletionItem[] = [
     insertText: "unzip ${1:path/to/archive}",
   },
   {
+    documentation: 'file characteristic filter',
+    label: 'find:filter-by-file-characteristic',
+    insertText: "find ${1:path/to/directory} -${2|readable,writable|}",
+  },
+  {
+    documentation: 'file type filter',
+    label: 'find:filter-by-file-type',
+    insertText: "find ${1:path/to/directory} -type ${2|b,c,d,p,f,l,s,D|}",
+  },
+  {
+    documentation: 'file name filter',
+    label: 'find:filter-by-file-name',
+    insertText: "find ${1:path/to/directory} -${2|name,iname|} ${3:path/to/file}",
+  },
+  {
+    documentation: 'condition combination',
+    label: 'find:filter-by-file-combined-condition',
+    insertText: "find ${1:path/to/directory} ${2:condition} -${3|and,or,a,o|} ${4:condition}",
+  },
+  {
+    documentation: 'modification',
+    label: 'find:modify-all',
+    insertText: "find ${1:path/to/directory} ${2:condition} -exec ${3:action} \;",
+  },
+  {
     documentation: 'home directory',
     label: '~',
     insertText: '$HOME',
--- a/server/src/snippets.ts
+++ b/server/src/snippets.ts
@@ -19,15 +19,14 @@
  * - for Bash parameter expansions it's '[<prefix>]<expression>', where:
  *   - <prefix> is one of 'set'/'error'
  *     used when expansion modifies variable or prints error to stderr
- *   - <expession> is 'if-(set|unset)[-or-[not-]null]'
+ *   - <expression> is 'if-(set|unset)[-or-[not-]null]'
  *   - term delimiter: dash, like 'set-if-unset-or-null'
  * - for Bash brace expansion it's 'range'
  * - for Bash documentation it's one of 'documentation'/'<documentation>'
  * - for Bash functions it's one of 'function'/'<function>'
  * - for Bash builtins it's '<builtin>'
  * - for Bash character classes it's '<character-class>'
- * - for Sed it's 'sed:<expression>'
- * - for Awk it's 'awk:<expression>'
+ * - for external command it's '<command>:<expression>'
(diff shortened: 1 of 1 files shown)


skovhus:
Thanks. There seems to be some formatting issues. Running yarn lint should fix it.

Shane-XB-Qian:
to be honest, those un-general chars in the label mostly would cause issues, 
i would suggest you gave up those non word chars;
or actually the snip had committed in this repo looks not very good quality,
how about rollback but just keep using snip from local plugin? that mostly maintained by community perhaps better?

EmilySeville7cfg:
Can you clarify whar quality issues do you mean and what extension you propose to use? I remember there was a shellman VS Code extension.

skovhus:
> how about rollback but just keep using snip from local plugin?

The argument for having the snippets as part of the language server is that the usage is much wider than an extension for vscode. Bash language server is used in all editors supporting the language server protocol (which is basically all major code editors or IDEs). 

But if quality is a concern, then we should fix that. I’m personally barely using snippets, so I’m unaware of the best practices here.

Shane-XB-Qian:
> The argument for having the snippets as part of the language server is that the usage is much wider than an extension for vscode.

maybe wider, but not certainly better.
actually it still required some additional local plugins of that editor supported, even server had snips, but if no or not supported well or config well, then those snips looks buggy or dup.
I am not talking about vscode always, actually I am not a vscode user.
if user really wants snips, local ones maybe more *flexible* and well maintained by community by more people. (though just a wish :smile, sooner or later you would find it perhaps became mess, it's not a coding language eventually, but just some snip)
for now I only found bashls and awkls had this which both PR from Emily.
anyway, it's your project, you call.

--
shane.xb.qian

skovhus:
Thanks for the additional context. I’m not sure I fully understand why specialized snippets for a single editor can be more flexible or better maintained than an LSP based on. 

Personally I’m not bothered by the bash language server snippets, but I’m curious if you would suggest having a config flag to disable them?

EmilySeville7cfg:
> or config well, then those snips looks buggy or dup.

Can you propose your suggestions on enhancing snippets? Somewhere, maybe as an issue, maybe as PR. I wanna see what you miss in snippets.

Shane-XB-Qian:
> I’m not sure I fully understand why specialized snippets for a single editor can be more flexible or better maintained than an LSP based on.

well, this is open source, basically it's just a json file (or similar), user can modify, remove, add or just delete it.
compare to snips in server, maybe you can add a flag to disable it, but well, it looks a like tech burden? or dup?

Shane-XB-Qian:
`"sed ${1|-E,--regexp-extended|} ':${2:x} N $! b$2 ${3:command}' ${4:file}"`  (from #923)

`$!` made following expanding failed, you may need to avoid or change to another way.
// looks same case if `$$`.

Shane-XB-Qian:
```patch
diff --git a/server/src/snippets.ts b/server/src/snippets.ts
index 2730fc1..ff3d491 100644
--- a/server/src/snippets.ts
+++ b/server/src/snippets.ts
@@ -594,13 +594,13 @@ export const SNIPPETS: BashCompletionItem[] = [
     insertText: "sed '' ${1:file}",
   },
   {
     documentation: 'file read',
     label: 'file-read',
     insertText:
-      "sed ${1|-E,--regexp-extended|} ':${2:x} N $! b$2 ${3:command}' ${4:file}",
+      "sed ${1|-E,--regexp-extended|} ':${2:x} N \\$! b$2 ${3:command}' ${4:file}",
   },
   {
     documentation: 'skip first',
     label: 'skip-first',
     insertText: 'tail ${1|-n,-c,--lines,--bytes|} +${2:count}',
   },
@@ -637,26 +637,26 @@ export const SNIPPETS: BashCompletionItem[] = [
   {
     documentation: 'device',
     label: 'device',
     insertText: '/dev/${1|null,stdin,stdout,stderr|}',
   },
   {
-    documentation: 'completion',
-    label: 'completion definition',
+    documentation: 'completion definition',
+    label: 'completion-definition',
     insertText: [
       '_$1_completions()',
       '{',
       '\treadarray -t COMPREPLY < <(compgen -W "-h --help -v --version" "\\${COMP_WORDS[1]}")',
       '}',
       '',
       'complete -F _$1_completions ${1:command}',
     ].join('\n'),
   },
   {
-    documentation: 'comment',
-    label: 'comment definition',
+    documentation: 'comment definition',
+    label: 'comment-definition',
     insertText: '# ${1:description}',
   },
 ].map((item) => ({
   ...item,
   documentation: {
     value: [
```

a bit complex or no idea some of them, 
but since this had been merged to lsp server side, then let's try to refine it a bit.

Shane-XB-Qian:
@EmilyGraceSeville7cf if you still wish to merge your left 3 snip PR (or reverted them & previous merged snip PR all?), maybe better tidy up rebase and squash them (and above my patch), and not include the dependency npm json/yaml files,  for now TBH a bit messy :smile: or there conflicted at least.

Shane-XB-Qian:
@EmilyGraceSeville7cf if you still wish to merge your left 3 snip PR (or reverted them & previous merged snip PR all?), maybe better tidy up and squash them (and above my patch), and not include the dependency npm json/yaml files,  for now TBH a bit messy :smile: or there conflicted at least.

PR #909, opened 2023-07-14
fix(man): Use cat as man pager

mikesmithgh:
Hi 👋, I ran into an issue when I configured Neovim as my default man pager via `export MANPAGER='nvim +Man!'`. bash-language-server is resolving commands with `man` but in my case `man` does not close because it is using Neovim. This can happen for any command in `MANPAGER` that doesn't exit. Below are a couple examples with different options:

### nvim 
man-nvim

### less
man-less

### cat
man-cat

### vim
man-vim

In this PR I am explicitly settings the pager to `cat` to avoid unexpected formatting.

Please let me know if you have any questions or suggestions. Thanks!

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Bash Language Server
 
+## 4.10.3
+
+- Use cat as man pager https://github.com/bash-lsp/bash-language-server/pull/909
+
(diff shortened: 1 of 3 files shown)


skovhus (approved):
Awesome, thanks for fixing this!

PR #913, opened 2023-07-16
Feature/allow different delimiters for sed and prettify labels

EmilySeville7cfg:
- allow `|` delimiter for sed commands
- use `-` instead of `:` in labels


skovhus:
Thanks!

> use - instead of : in labels

Curious what the reason behind this change. Is that more standard?

EmilySeville7cfg:
> Curious what the reason behind this change. Is that more standard?

This change reflects [this](https://github.com/bash-lsp/bash-language-server/pull/897#issuecomment-1623641500) comment. Actually there is no standard here AFAIK. You can meet [spaces](https://github.com/yousefvand/shellman/blob/master/snippets/snippets.json#L4) in labels too.

I am thinking about replacing `-` with spaces. But I don't know whether Vim/Emacs will be okay with such snippets. @Shane-XB-Qian, do you know anything about this? Is it fine to have spaces for snippets used in Vim/Emacs?

Shane-XB-Qian:
> I am thinking about replacing `-` with spaces. But I don't know whether Vim/Emacs will be okay with such snippets. @Shane-XB-Qian, do you know anything about this? Is it fine to have spaces for snippets used in Vim/Emacs?

i preferred `_` (underscore) (though `-` maybe workable) IF snips have to store in server.

-- 
shane.xb.qian

EmilySeville7cfg:
This PR is closed in favor to #923.

PR #945, opened 2023-08-18
Support colors and text attributes

EmilyGraceSeville7cf:
Motivation is to reduce the need to google color codes each time.


EmilySeville7cfg:
I suggest merge this PR with #937 at once. :)

Issue #1024, opened 2023-11-06
Question: what is meant by "We strongly recommend that you install shellcheck to enable linting: https://github.com/koalaman/shellcheck#installing"?


jerabaul29:
What is meant by "We strongly recommend that you install [shellcheck](https://www.shellcheck.net/) to enable linting: https://github.com/koalaman/shellcheck#installing" in the readme?

Does it mean that one should install spellcheck and then use it "on its own"?

Or does it mean that by i) installing spellcheck, ii) installing bash-language-server, then bash-language-server will run spellcheck and provide spellcheck output through the bash-language-server LSP? If so, how often / when is spellcheck triggered? (on save? on other events?).

Could a short explanation about this be added to the readme? :)

skovhus:
Yes, we just find the binary and run it from the LSP server. PRs clarifying this is more than welcome.

skovhus:
Whenever a file changes:

https://github.com/bash-lsp/bash-language-server/blob/main/server/src/server.ts#L154

But we debounce the request for 500 ms.

PR #1027, opened 2023-11-06
Explain shellcheck dependency; add snap install note

jerabaul29:
As mentioned in #1024 , explain that shellcheck is a dependency to get linting / code analysis, that is used by bash-language-server to generate diagnostics etc.

Also add a couple of minor readme updates (mention snap install on Ubuntu).


skovhus (approved):

skovhus:
Closes https://github.com/bash-lsp/bash-language-server/issues/1024

PR #1052, opened 2023-11-28
Add --help documentation source

SuperCuber:
Some commands are installed without a manpage (for example `pip` and stuff installed by it) and therefore neither `man` nor `help` work on them. This adds a fallback to `command --help` for those.

There is a concern here - some commands do not care about `--help` or any flags at all and will happily take it as a positional argument, and might have side effects - like `echo`. These are a small minority though and hopefully most of them have a manpage.

I've tested it manually and it looks good. If I should write tests for this let me know :)


ChillerDragon:
[teeworlds](https://github.com/teeworlds/teeworlds) and [ddnet](https://github.com/ddnet/ddnet) ignore --help and start as graphical applications. In general this approach seems dangerous. A user could have any kind of self rolled script that ignores --help and does something destructive.

me-mariosteve:
The help flag is not and was never a standard. The behaviour of a random program can not be predicted. The "side effects" are not acceptable at all. The "commands" that do not care about that flag are not a "minority". The fact that someone merged this without further discussion scares me.

SuperCuber:
@me-mariosteve 
> The "commands" that do not care about that flag are not a "minority"

This was a faulty assumption on my end and hopefully this change gets undone soon 🙃 

> The fact that someone merged this without further discussion scares me.

The state of OSS should scare everyone. Millions of lines of code are running on your machine (from libraries, plugins, etc) that are vetted by a tiny amount of people (usually one or two), most of them you are able to pull and run without verification as soon as it was pushed by the maintainer.
The fact that there are "only" a couple large-scale security incidents per year arising from this situation is a miracle.

Issue #933, opened 2023-08-06
Code actions to disable shellcheck rules for lines or entire files


polyzen:
### What is the problem this feature will solve?

Currently it seems there are only code actions to apply fixes when available. Code actions to disable the rules (using comments) would also be appreciated.

https://github.com/bash-lsp/bash-language-server/issues/490#issuecomment-1406298247

### What is the feature you are proposing to solve the problem?

Additional code actions, eg. null-ls offers:
```
Disable ShellCheck rule 2059 for the entire file
Disable ShellCheck rule 2059 for this line
```

xero:
+1 for this idea 💡

@polyzen i’m not sure what editor you use, but as a stopgap, i’ve written a little neovim lua command to insert a shellcheck disable statement based on the current line’s lsp diagnostics into your buffer:

https://github.com/xero/dotfiles/commit/f3821bebb4ec3149a83413c606140741124c45f7#diff-cf8a96b3f61615a85a1fb4313e82a75b58d6e0bd71869e9c3b1ead9d1bbd91ab

hope this helps for now.

Zeioth:
linting should be a responsability of a linter, and not a lsp server, but if we are doing this, +1.

Zeioth:
@xero I think I found a better solution for neovim: https://github.com/bash-lsp/bash-language-server/issues/999#issuecomment-1763539911

![screenshot_2023-10-16_01-48-35_133417388](https://github.com/bash-lsp/bash-language-server/assets/3357792/a9ad863c-40dc-4a1d-9877-fa87754c1463)

ThomasFaivre:
Hello there!

I have made a first draft to try to address this issue:
https://github.com/ThomasFaivre/bash-language-server/commit/d0b80ffaa580b25e91ed1f996ea258708f83ebdb
It does stuff (tested in neovim+LazyVim), but it is not finished for several reasons:

- [ ] handle indentation
- [ ] handle existing disable lines
- [ ] make tests
- [ ] Stuff I'm missing

I wanted to get some inputs before continuing.
And I have some questions. I am very new to Typescript (like, this is the second time I am touching javascript related code :D ), so I apologize if my questions are strange.

- What would be the best way to get indentation? I'm assuming it would have something to do with the `TextDocument`, but how to get a given line? Kind of the same question to detect an existing disabling line.
[...]

PR #1069, opened 2023-12-22
shellcheck: add code actions to disable rule

ThomasFaivre:
Following my [comment in #933](https://github.com/bash-lsp/bash-language-server/issues/933#issuecomment-1858657810), I ended up finishing my draft.
It is now fully functionnal:
* [x]  handle indentation
* [x]  handle existing directives
* [x]  make tests
* [x]  handle multiline commands
* [x]  insert sorted disable directives
* [ ]  Stuff I'm missing

Tested with Neovim (NVIM v0.10.0-dev-1846+gf31f260f0).

Therefore, I would like your opinion on the implementation.
I have already spotted some room for improvements:

- [ ] use directive.ts existing module
- [ ] use tree sitter (multiline detection, indentation maybe?, disable whole scope?)
- [ ] split code into different modules
- [ ] some doc on public methods. Maybe on private as well?
- [ ] disable all errors of a given line (for the line and for the whole file)

What do you think?
Thanks!

### DIFF
--- a/server/src/__tests__/server.test.ts
+++ b/server/src/__tests__/server.test.ts
@@ -222,7 +222,7 @@ describe('server', () => {
         {} as any,
       )
 
-      expect(result).toHaveLength(1)
+      expect(result).toHaveLength(3)
       const codeAction = (result as CodeAction[])[0]
       expect(codeAction.diagnostics).toEqual([fixableDiagnostic])
       expect(codeAction.diagnostics).toEqual([fixableDiagnostic])
--- a/server/src/shellcheck/index.ts
+++ b/server/src/shellcheck/index.ts
@@ -7,7 +7,7 @@ import { TextDocument } from 'vscode-languageserver-textdocument'
 
 import { debounce } from '../util/async'
 import { logger } from '../util/logger'
-import { analyzeShebang } from '../util/shebang'
+import { analyzeShebang, getShebang } from '../util/shebang'
 import { CODE_TO_TAGS, LEVEL_TO_SEVERITY } from './config'
 import {
   ShellCheckComment,
@@ -25,7 +25,7 @@ type LinterOptions = {
 
 export type LintingResult = {
   diagnostics: LSP.Diagnostic[]
-  codeActions: Record<string, LSP.CodeAction | undefined>
+  codeActions: Record<string, LSP.CodeAction[]>
 }
 
 export class Linter {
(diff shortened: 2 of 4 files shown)


server/src/__tests__/server.test.ts
```diff
@@ -222,7 +222,7 @@ describe('server', () => {
         {} as any,
       )
 
-      expect(result).toHaveLength(1)
+      expect(result).toHaveLength(3)
```
skovhus, line 225: Let us check what all the code actions are about. Consider just doing a snapshot test of the entire result object.

server/src/shellcheck/index.ts
```diff
@@ -180,7 +180,15 @@ export class Linter {
...
 }
 
-function mapShellCheckResult({ uri, result }: { uri: string; result: ShellCheckResult }) {
+function mapShellCheckResult({
+  document,
+  uri,
+  result,
+}: {
+  document: TextDocument
+  uri: string
```
skovhus, line 189: No reason to pass in the uri if we have the document.

server/src/shellcheck/index.ts
```diff
@@ -237,6 +267,242 @@ function mapShellCheckResult({ uri, result }: { uri: string; result: ShellCheckR
  * Copyright (c) Timon Wong
  */
 class CodeActionProvider {
+  private static getLine({
+    document,
+    lineNb,
+  }: {
+    document: TextDocument
+    lineNb: number
```
skovhus, line 275: Nit: the `nb` abbreviation and postfix seems unnecessary.

server/src/shellcheck/index.ts
```diff
@@ -237,6 +267,242 @@ function mapShellCheckResult({ uri, result }: { uri: string; result: ShellCheckR
...
+    if (!directives) {
+      // insert new directive line
+      const offset = insertBefore ? 0 : 1
+      return this.getInsertLineTextEdit({
+        lineNb: directiveLineNb + offset,
+        newText: `${indentation}# shellcheck disable=SC${code}\n`,
+      })
+    }
+
+    // TODO: use directives module
```
skovhus, line 342: 👍

server/src/shellcheck/index.ts
```diff
@@ -237,6 +267,242 @@ function mapShellCheckResult({ uri, result }: { uri: string; result: ShellCheckR
...
+      newText: directiveLine.replace(
+        DISABLE_REGEX,
+        `disable=${codes.map((code) => `SC${code}`).join(',')}`,
+      ),
+    })
+  }
+
+  private static getIndentation(line: string): string {
+    const match = line.match(/^([\s]*)/)
+    if (match && match[1]) {
```
skovhus, line 368: Probably worth exploring using the AST for this.

skovhus (commented):
Thanks! Seems like a great starting point. I suggest adding some unit test and the. After that see if we should improve the implementation.

ambroisie:
This should be closed, as it has been superseded by #1449.

Issue #1064, opened 2023-12-07
shellcheck: forcing to work with zsh does not work


chrisgrieser:
### Code editor

nvim

### Platform

macOS 14.1 (arm)

### Version

5.0.0

### What steps will reproduce the bug?

shell file with zsh shebang and an issue that shellcheck will reprot:

```bash
#!/usr/bin/env bash
cat foo | grep "foobar"
# here shellcheck reports an issue
```

```bash
#!/usr/bin/env zsh
cat foo | grep "foobar"
# here it does not
```

### How often does it reproduce? Is there a required condition?

always

### What is the expected behavior?

When using shellcheck, you can actually force it to check zsh files by adding `--shell=bash`. While not perfect, it's still useful.

However, when passing `bashIde.shellcheckArguments = "--shell=bash"`, bashls still refuses to lint files with a zsh shebang. I assume there is something in the bashls source-code which stops shellcheck from working on zsh files, even if it nromally works in shellcheck.

### What do you see instead?

shellcheck not working as soon as you add a zsh-shebang

### Additional information

- related issue: #929
- potential cause? #536

skovhus:
Contributions are more than welcome.

0xdevalias:
Context for the close/reopen:

> skovhus Sorry to bother you again, but it seems this PR did not fix the issue.
> 
> The problem described in #1064 still occurs, and shellcheck does not become active in a zsh file, even if passing `bashIde.shellcheckArguments = "--shell=bash"`. I assume the problem is that the LS somehow still does not enable shellcheck until it is a bash file?
> 
> _Originally posted by @chrisgrieser in https://github.com/bash-lsp/bash-language-server/issues/1133#issuecomment-2569158675_

PR #1081, opened 2024-01-11
feat: handle shellcheck's `shell` directive

### LINKED ISSUE #929: The "shell" directive has no effect in a shell script file
### Code editor

Sublime Text

### Platform

Windows, Linux

### Version

5.0.0

The following script raises a shellcheck warning on the "set" line:
```shell
#!/bin/sh
# shellcheck shell=ash
set -euxo pipefail
```

However it should not because I explicitly defined the shell to eliminate the warning.

Always

I would expect the shell directive to be taken into account, so no warning should appear on the "set" line

### What do you see instead?

The shell directive is ignored

### Additional information

I  suspect that the problem is caused by the LSP language server passing the `--shell` option to the command-line which have the priority over the shell directive defined in my script file. But, AFAIK, there's no way to change/remove the value of the `--shell` option, right?

liljaylj:
Fixes #929 #1064


skovhus (approved):
Thanks!

PR #1086, opened 2024-01-21
feat: support leading dynamic section when sourcing

### LINKED ISSUE #659: Sourcing: support basic variable expansions
https://github.com/bash-lsp/bash-language-server/pull/244 introduced sourcing aware symbol resolution where source commands (e.g. `. myfile.sh` or `source ./file.bash`) is used for figure out which symbols are available in a given context. 

The current [implementation](https://github.com/bash-lsp/bash-language-server/blob/main/server/src/util/sourcing.ts) is fairly crude and can be extended a bit to resolve basic variable expansions and sourcing statements in functions and loops.

### LINKED ISSUE #926: Sourcing: "strip one leading, dynamic section before trying to locate the rest"
### What is the problem this feature will solve?

When writing code like this:
```shell
here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
. "$here/config.sh"
. "$here/utils.sh"
```
shellcheck will understand that it should search the script dir for the files, but bash lsp doesn't.

### What is the feature you are proposing to solve the problem?

```
To support the common pattern of . "$CONFIGDIR/mylib.sh", ShellCheck strips one leading, dynamic section before trying to locate the rest. That means that ShellCheck will look for config.sh and utils.sh in the same directory as this script when it is being checked:
```
So, let's `strip one leading, dynamic section before trying to locate the rest` :smile:

This can work with #811 by only taking effect when the variable is truly dynamic, instead using the value if it's constant (ie set earlier in the script)

### What alternatives have you considered?

`# shellcheck source=./.utils.d/output.sh` - this gets quite frustrating when shellcheck understands the location of the file, but bash-lsp doesn't

aarondill:
This does the same thing that ShellCheck does ([here](https://github.com/koalaman/shellcheck/blob/ba86c6363c30a5dbefd0b8b9a7c5f4ab0478dc91/src/ShellCheck/Parser.hs#L2277-L2283)).

`To support the common pattern of . "$CONFIGDIR/mylib.sh", ShellCheck strips one leading, dynamic section before trying to locate the rest.`

fixes #926, fixes #659

### DIFF
--- a/server/src/__tests__/analyzer.test.ts
+++ b/server/src/__tests__/analyzer.test.ts
@@ -123,6 +123,28 @@ describe('analyze', () => {
           "severity": 3,
           "source": "bash-language-server",
         },
+        {
+          "message": "Source command could not be analyzed: failed to resolve path.
+
+      Consider adding a ShellCheck directive above this line to fix or ignore this:
+      # shellcheck source=/my-file.sh # specify the file to source
+      # shellcheck source-path=my_script_folder # specify the folder to search in
+      # shellcheck source=/dev/null # to ignore the error
+
+      Disable this message by changing the configuration option "enableSourceErrorDiagnostics"",
+          "range": {
+            "end": {
+              "character": 49,
+              "line": 26,
+            },
+            "start": {
+              "character": 0,
+              "line": 26,
+            },
+          },
+          "severity": 3,
+          "source": "bash-language-server",
+        },
       ]
     `)
 
@@ -847,6 +869,7 @@ describe('initiateBackgroundAnalysis', () => {
       [expect.stringContaining('parse-problems.sh: syntax error')],
       [expect.stringContaining('sourcing.sh line 16: failed to resolve path')],
       [expect.stringContaining('sourcing.sh line 21: non-constant source not supported')],
+      [expect.stringContaining('sourcing.sh line 26: failed to resolve path')],
     ])
 
     // Intro, stats on glob, one file skipped due to shebang, and outro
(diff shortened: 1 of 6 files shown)


aarondill:
<details>
<summary>this can now handle all the below cases:</summary>

```shell
. "$dir/file"
. "$dir/other/file"
. "$dir"/file
. $dir/file
. $dir"/file"
. "$dir"/"file"
. "${dir:-.}"/"file"
. ${dir}/"fi"l'e'
source "$dir/file"
```
</details>

<details>
<summary>It intentionally fails on these:</summary>

```shell
. "$dir/$file"
. "$dir/other/$file"
. "$(cmd)/file"
. $(cmd)/file
```
</details>
Tests are coming soon! (once i figure out this testing framework)

Shane-XB-Qian:
> It intentionally fails on these:

is there a way to make it work? seems that's the much more usage case.

aarondill:
> > It intentionally fails on these:
> 
> 
> 
> is there a way to make it work? seems that's the much more usage case.

potentially, but these cases are process substitution and dynamic *file* sourcing. if you really want to use process substitution, assign it to a variable, then use that - this way it will work with shellcheck too. as to dynamic file sourcing, at the point this is evaluated, there's no way to get the "static" value of a variable, so there's no meaningful way to support that.

the current implementation matches shellcheck's behavior

Shane-XB-Qian:
> there's no way

:thinking:

isc-aray:
This would be great! All of our sources are of the form `. "$ScriptDir/something.sh"` or `. "$ScriptDir/dir/something.sh"`.

skovhus:
Sorry for not getting back to you before, I'm not actively maintaining this project TBH. But the change here looks really useful.

skovhus:
@aarondill seems to one failing case, do you mind having a look?

aarondill:
I squashed it (mostly to make rebasing easier) and rebased to `main`. After adding a check to avoid a regression of f8b202b, it passes all the tests!

PR #1120, opened 2024-03-21
Fix pnpm add flags

Jimmyscene:
It looks like when the repository was refactored from yarn-> pnpm, the --dev flag was not updated properly.

This causes the auto update of treesitter to fail(https://github.com/bash-lsp/bash-language-server/pull/1115)

https://pnpm.io/cli/add#--save-dev--d


skovhus (approved):

skovhus:
Thanks!

PR #1133, opened 2024-03-30
fix: do not overwrite user-provided shellcheck `--shell` (#1064)

### LINKED ISSUE #1064: shellcheck: forcing to work with zsh does not work

chrisgrieser:
This is an attempt to fix #1064. I am really not familiar with the inner working of LSPs, but the fix appears to be simple enough, if I am not mistaken here.


skovhus:
This PR should be an okay workaround, but we should already infer the shell name based on the file name and the shebang. Can you come with an example file where this doesn't work as intended?

chrisgrieser:
> we should already infer the shell name based on the file name and the shebang

That's exactly the problem. If the file is inserted to be a zsh file, `--shell=zsh` gets passed which stops shellcheck from working. 

To force shellcheck to euro with zsh files, you need to explicitly pass `--shell=bash`, which the bash ls currently does not allow. Thus this PR.

skovhus:
Thanks, that makes sense. I see that we actually fall back to using `bash` if we cannot infer the dialect. We could consider doing the same for zsh files, although your approach is safer.

skovhus (approved):

skovhus:
Sorry for the delay here...

chrisgrieser:
@skovhus Sorry to bother you again, but it seems this PR did not fully fix the issue and I cannot really figure out why.

The problem described in #1064 does still occur, and shellcheck does not become active in a zsh file, even if passing `bashIde.shellcheckArguments = "--shell=bash"`. I assume the problem lies now does not lie with the wrong shellcheck arg, but the LS not enabling shellcheck until it is a bash file?

PR #1135, opened 2024-03-30
clear diagnostics on close document

bzy-debug:
Currently, bash-language-server does not clear diagnostics when user close an document. This causes bad user experience.

This PR fixes this problem, it makes sure that server clear diagnostics when document was closed


Shane-XB-Qian:
curious to ask: how such 'clear' helped? 
just send a empty `[]` to client? what's the different if just do nothing when closing? 
or was it helpless but adding a bit burden?

bzy-debug:
Quote from lsp [specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics)

> Diagnostics are “owned” by the server so it is the server’s responsibility to clear them if necessary. The following rule is used for VS Code servers that generate diagnostics:
> - **if a language is single file only (for example HTML) then diagnostics are cleared by the server when the file is closed.** Please note that open / close events don’t necessarily reflect what the user sees in the user interface. These events are ownership events. So with the current version of the specification it is possible that problems are not cleared although the file is not visible in the user interface since the client has not closed the file yet.
> - if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache).

clear here usually means send empty diagnostics (i.e. `[]`) to client, so that client's UI won't show any diagnostics on this file.

AFAIK, bash-lsp currently does not have very much project system concept, so it's resonable to clear diagnostics.

If server does not clear diagnostics, the diagnostics will live forever in client's UI, even after user has deleted the file, util user restart the extension host. This is kind of annoying, in my opinon.

Here's a video illustration:


https://github.com/bash-lsp/bash-language-server/assets/71200607/2a2c150e-d454-47b1-92e0-1f5ff52abdab

Shane-XB-Qian:
the problem or question is send [] to client seems useless since/if that file had been closed or even burden since hanlder needed to ACK it.
perhaps other kinds `clear` which server side only then maybe helpful/useful.
// I cannot see/open your video for now.

--
shane.xb.qian

bzy-debug:
Does sending [] to client not clear diagnostics on your side? That works well on my side.

Here's a illustration: (I'm using vscode version 1.89.0 with bash-lsp on main 81125b8b2989abe924145a2f7cdae3aaf5615d4a)


https://github.com/bash-lsp/bash-language-server/assets/71200607/2f9a81e9-9de1-4c74-927a-a47863b7fc9d

P.S. this clear diagnostics behavior might be different among different editors, since they implement language client differently. If there are any cross-client compatible methods to clear diagnostics, i'd like to hear.

Shane-XB-Qian:
i am seeing your video now, i think it's probably not a problem,
i am not a vsc user, but i guess it's its same behavior for others langs.
// sending a fake `[]` to client i supposed it is not a `clear`, or if you really like to `hide` it then it maybe a just client option.

Shane-XB-Qian:
@skovhus so did you plan to keep this, or actually it should be a vscode option to hide it (if there was one) ?

skovhus:
I haven’t seen this causing any harm. Have you?

Issue #1134, opened 2024-03-30
`'textDocument/rename'` does not include variable references in `(( ))` (arithmetic)


chrisgrieser:
### Code editor

nvim

### Platform

macOS 14.1 (arm)

### Version

5.1.2

### What steps will reproduce the bug?

create a file with this code:

```bash
foo=123
foo=$((foo + 1))
echo "$foo"
```

renaming `foo` to `bar` results in this:

```bash
bar=123
bar=$((foo + 1))
echo "$bar"
```

### What is the expected behavior?

References in a double bracket `(())` should be affected by the renaming.

### What do you see instead?

No reference in a double bracket `(())` is affected by the renaming.

mcecode:
Hi. I'm the one who implemented rename symbol in #915.

One of the limitations I bumped into was that there are cases when variables are not typed properly as variables by the parser used (tree-sitter-bash) and therefore aren't picked up for renaming. Unfortunately, that's the case with this issue, the variable `foo` inside the arithmetic expansion isn't labelled by the parser as a variable and so it isn't renamed.

However, as indicated in #943, there's been a lot of fixes done recently to tree-sitter-bash and when trying out your code sample in the [tree-sitter playground](https://tree-sitter.github.io/tree-sitter/playground), `foo` in the arithmetic expansion is correctly typed as a variable. Currently, when looking at the [`parser.info` file](https://github.com/bash-lsp/bash-language-server/blob/f78058fdcf8690c62f85f2d030e445c0a42a72ad/server/parser.info#L1) , it seems that we're using an old version of tree-sitter-bash (0.19.0) from 2022. I think this issue should be resolved when tree-sitter-bash is updated to the latest version (0.21.0), though I'm not sure if there are any blockers in doing that since #911 seems to indicate that there was a problem with upgrading.

Issue #943, opened 2023-08-18
tree-sitter-bash has updated to 0.20.0 w/ many bug fixes


amaanq:
Hey, I've seen in some issues bash-ls uses tree-sitter-bash under the hood for parsing.

I've recently fixed many (20+?) issues upstream in tree-sitter-bash, and parsing works much better for many cases 

I'm just dropping this by so you guys can update and try it out, and hopefully improve the ls :)

skovhus:
Thanks. Yeah we have actually been using a custom build of tree sitter for some time based on their main branch. 

But I’m happy if we can get rid of those GitHub actions and start relying on packages being published again. 🤞

PR #1148, opened 2024-04-21
Upgrade dependencies

Jimmyscene:
Upgrade web-tree-sitter, tree-sitter-cli and tree-sitter-bash

These sort of all need to be done at the same time else everything blows up

Update rootNode.hasError() to hasError per
https://github.com/tree-sitter/tree-sitter/pull/3103/files

Update build-wasm flag to build --wasm per
https://github.com/tree-sitter/tree-sitter/commit/99b93d83a11a84ad9a23f5248e126fe45c1a61d3


server/src/util/__tests__/__snapshots__/sourcing.test.ts.snap
```diff
@@ -126,7 +126,7 @@ exports[`getSourcedUris returns a set of sourced files (but ignores some unhandl
         "line": 25,
       },
     },
-    "uri": "file:///Users/bash/issue206.sh",
+    "uri": null,
```
skovhus, line 129: What happened here. Cannot be parsed anymore?

Jimmyscene: Sorry, marked this as draft hoping it wouldn't bother you until this was ready. That happened automatically when I updated the snapshots, I'm still poking to see what happened exactly

skovhus: No worries. And thanks for looking into this.

server/src/__tests__/server.test.ts
```diff
@@ -1740,7 +1740,7 @@ describe('server', () => {
     describe('Edge or not covered cases', () => {
       it('only includes variables typed as variable_name', async () => {
         const iRanges = await getFirstChangeRanges(getRenameRequestResult(106, 4))
-        // This should be 6 if all instances within let and arithmetic
+        // This should be 6 if all instances within let and c-style for loop
```
Jimmyscene, line 1743: As I note in the commit message,

The anazlyer is now detecting variables within arithmetic expressions,
but still not let or c style for loops

So a rename on

```
for i in 1 2 3; do
	echo "$i"
done

let i=0
for (( ; i < 10; i++)); do
	echo "$((2 * i))"
done
```

Would be changed to

```
for new_name in 1 2 3; do
	echo "$new_name"
done

let i=0
for (( ; i < 10; i++)); do
	echo "$((2 * new_name))"
done
```

Which is definitely wrong, but I'm interested to hear opinions from the maintainers on this

server/src/util/__tests__/__snapshots__/sourcing.test.ts.snap
Jimmyscene: Closing this convo since I undid this change, and marking this as ready for review. Thanks for your time!

server/src/__tests__/server.test.ts
skovhus: @mcecode what do you think about this?

mcecode: I think this is to be expected since, as far as I can tell, tree-sitter-bash only updated the types of variables from `word` to `variable_name` within arithmetic expressions. For those other constructs they're still typed as `word`. The current implementation of rename symbol only renames variables typed as `variable_name`.

mcecode:
Since I was tagged, I'll leave some thoughts/notes here.

If/once this gets merged,  #1134 can be closed as per my comment there https://github.com/bash-lsp/bash-language-server/issues/1134#issuecomment-2040637632.

#943 can also be closed since that's just urging us to upgrade. On that note, I think tree-sitter-bash can be installed as is from the npm registry now instead of the latest from GitHub since they've started shipping stable versions there again, the latest being 0.21.

skovhus:
Thanks @Jimmyscene – and thank you @mcecode for adding valuable comments.

PR #1155, opened 2024-04-25
Update release-client.sh script

Jimmyscene:
Update vscde to latest version (vsce is depresated in favor of @vscode/vsce) 

The current code has an `|| echo` that masks failures from vsce publish, in order to not fail the pipeline when the published version already exists.

The latest version of vsce adds a --skip-duplicate flag, this allows us to fail silently if the version already exists on the marketplace, so IMO this is a better way to handle that

### DIFF
--- a/server/src/__tests__/__snapshots__/server.test.ts.snap
+++ b/server/src/__tests__/__snapshots__/server.test.ts.snap
@@ -653,11 +653,11 @@ exports[`server onRenameRequest Workspace-wide rename returns correct WorkspaceE
         "newText": "newName",
         "range": {
           "end": {
-            "character": 53,
+            "character": 10,
             "line": 16,
           },
           "start": {
-            "character": 43,
+            "character": 0,
             "line": 16,
           },
         },
--- a/scripts/release-client.sh
+++ b/scripts/release-client.sh
@@ -13,5 +13,5 @@ pnpm verify:bail
 
 cd vscode-client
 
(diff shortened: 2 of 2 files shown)


Jimmyscene:
I don't actually have a way to test this, but I think it should be okay

If you disagree with this, feel free to close, but I thought it might be a good idea 😄 

Thanks for reviewing!

Jimmyscene:
(Marking as WIP while I fix the tests)

server/src/__tests__/__snapshots__/server.test.ts.snap
```diff
@@ -653,11 +653,11 @@ exports[`server onRenameRequest Workspace-wide rename returns correct WorkspaceE
         "newText": "newName",
         "range": {
           "end": {
-            "character": 53,
+            "character": 10,
```
skovhus, line 656: Hm, why did these change? 🤔

Jimmyscene: It’s because i moved tagRelease to a separate line, but removed one line above it (the comment) so the line number stays the same

skovhus: Ah, pretty nasty that I use the release script as a test file. My bad.

Jimmyscene: No worries, it's not really supposed to change that often haha

PR #1158, opened 2024-05-03
Install shellcheck and shfmt for deploy workflow

chris-reeves:
The merge of my `shfmt` changes broke the `deploy` workflow. Both of the release scripts call `pnpm verify:bail` (which runs the test suite), so these test dependencies need to be installed here as well.

Although `shellcheck` is included in GitHub's `ubuntu-latest` build image, we explicitly install it here both for consistency with the `verify` workflow and as a guard against this being removed from the build image in future.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,6 +11,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install shellcheck and shfmt (used for testing)
+        run: sudo apt-get install -y shellcheck shfmt
+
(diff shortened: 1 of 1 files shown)


skovhus:
Thanks for fixing!

chris-reeves:
@skovhus Don't know if you've noticed - the deploy is still failing, but this time because your access token has expired. Can't help with that one. :-)

David-Else:
Latest version I can install is 5.2.0, NPM needs sorting: https://www.npmjs.com/package/bash-language-server?activeTab=versions

PR #1171, opened 2024-06-07
Add .editorconfig support for shfmt

chris-reeves:
This PR adds support for reading `shfmt` config via `.editorconfig`.

If any `shfmt`-specific configuration properties are found in `.editorconfig` then the config in `.editorconfig` will be used and the language server config will be ignored. This follows `shfmt`'s approach of using either `.editorconfig` or command line flags, but not both. Note that only `shfmt`-specific configuration properties are read from `.editorconfig` - indentation preferences are still provided by the editor, so to format using the indentation specified in `.editorconfig` make sure your editor is also configured to read `.editorconfig`.

It is possible to disable `.editorconfig` support and always use the language server config by setting the "Ignore Editorconfig" configuration variable.

See #1165 for some discussion of the approach to this.


chris-reeves:
Fixed one of the tests to work with the older version of `shfmt` available on the GitHub runners.

skovhus (approved):

PR #1196, opened 2024-09-21
Restructure environmentVariablesUsed detection to avoid falsely alerting

blast-hardcheese:
Avoid incorrectly warning the user of `shfmt` envvar configuration:

```
03:53:00.510 WARNING ⛔️ Environment variable configuration is being deprecated, please use workspace configuration. The following environment variables were used: shfmt
```

Thank you!

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -101,10 +101,20 @@ export function getConfigFromEnvironmentVariables(): {
     },
   }
 
-  const environmentVariablesUsed = Object.entries(rawConfig)
-    .map(([key, value]) => (typeof value !== 'undefined' ? key : null))
-    .filter((key): key is string => key !== null)
-    .filter((key) => key !== 'logLevel') // logLevel is a special case that we ignore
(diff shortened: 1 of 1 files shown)


server/src/config.ts
```diff
@@ -101,10 +101,20 @@ export function getConfigFromEnvironmentVariables(): {
     },
   }
 
-  const environmentVariablesUsed = Object.entries(rawConfig)
-    .map(([key, value]) => (typeof value !== 'undefined' ? key : null))
-    .filter((key): key is string => key !== null)
-    .filter((key) => key !== 'logLevel') // logLevel is a special case that we ignore
```
skovhus, line 107: Why don't we filter out `shfmt` here like we already filter out `loglevel`?

blast-hardcheese: Personal preference, if I'm honest. I trend towards exhaustivity checks, where possible, but I'm happy to switch it back to the way it was and add a test to the filter.

skovhus:
Seems fixed on main

PR #1221, opened 2024-11-10
Support renaming variables within `read` commands

### LINKED ISSUE #1183: Renaming a variable changes all instances except those made on `read`
### Code editor

Neovim

### Platform

Linux

### Version

5.4.0

Any code where a variable is set and then potentially reset by the `read` command. For example, on this snippet:
```bash
CASE="$2"

# Check for arguments, if we didn't receive them, ask for them
if [[ -z $CASE ]]; then
	read -rp "Enter case number (must be 8-10 numerals): " CASE
fi
TARGETDIR="/srdata/$CASE"
```
Renaming `CASE` will change it everywhere except for the instance on `read -rp "Enter case number (must be 8-10 numerals): " CASE`.

Can be reproduced all the time.

The variable should be renamed everywhere.

### What do you see instead?

Rename skips instances of the variable used with the `read` command.

### Additional information

There are other instances of variables being used without the usual `$`. For example, you can omit the `$` inside arithmetic `$((...))`. I have not tested if these instances are also skipped by the rename.

mcecode:
Fixes #1183


skovhus (approved):
Sorry for missing this, it looks great!

PR #1227, opened 2024-11-19
fix(logs): prevent shfmt environment variable deprecation warning

### LINKED ISSUE #1226: Warrning: Environment variable configuration is being deprecated
### Code editor

nvim

### Platform

Linux (Ubuntu)

### Version

5.4.2

I don't have any of the environment variables, but opening a shell script logs the following warning to `~/.local/state/nvim/lsp.log` when the LSP server attaches to a buffer.

```
[WARN][2024-11-18 21:19:33] ...lsp/handlers.lua:625	"05:19:33.609 WARNING ⛔️ Environment variable configuration is being deprecated, please use workspace configuration. The following environment variables were used: shfmt"
```

It reproduces when the LSP server attaches to a buffer for the first time in the session. Closing neovim and reopening the file will add another warning to the log file.

I expect that the code correctly determines whether any of the config environment variables are being used.

### What do you see instead?

This might be the place the warning is talking about, because the `shfmt` key is an object.

https://github.com/bash-lsp/bash-language-server/blob/1ebe533c799fdc5833a72b0e8707384219acc964/server/src/config.ts#L104-L107

### Additional information

_No response_

benelan:
Prevent consistent warnings in the log file about deprecated shfmt environment variables. The change also slightly improves performance by reducing the iterations. 

Alternatively, I could check the `shfmt` object for environment variable usage instead of skipping them like `logLevel`. Let me know what you'd prefer!

Fixes #1226

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -102,9 +102,12 @@ export function getConfigFromEnvironmentVariables(): {
   }
 
   const environmentVariablesUsed = Object.entries(rawConfig)
-    .map(([key, value]) => (typeof value !== 'undefined' ? key : null))
-    .filter((key): key is string => key !== null)
-    .filter((key) => key !== 'logLevel') // logLevel is a special case that we ignore
+    .filter(
+      ([key, value]) =>
+        !['undefined', 'object'].includes(typeof value) &&
+        ![null, 'logLevel'].includes(key),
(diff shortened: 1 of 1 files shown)


skovhus (approved):
Thanks!

PR #1367, opened 2026-01-15
feat: better shell type detection

intelfx:
This PR improves bash-language-server behavior around unrecognized shell types, such as busybox sh. Currently, bash-language-server does not recognize busybox shell, but it also actively prevents shellcheck from recognizing it either.

Even if user specifices one directly via a `shellcheck shell=busybox` directive, we reject that directive because we do not know this dialect. Conversely, if the user does not specify a dialect in hopes of letting `shellcheck` figure it out, bash-language-server will implicitly pass a "guessed" dialect to shellcheck via a command line option, overriding its auto-detection.

1. For immediate effect, add and recognize a new "busybox" shell type. This ie needed to be able to pass it to shellcheck, even if we do not act on it.
2. In general, rework shell detection code such whenever a new shell dialect ever appears in future, we could handle it more gracefully even without recognizing it explicitly.

### DIFF
--- a/server/src/shellcheck/config.ts
+++ b/server/src/shellcheck/config.ts
@@ -10,6 +10,7 @@ export const SHELLCHECK_DIALECTS = [
   'bash',
   'dash',
   'ksh',
+  'busybox',
 ]
 
 // https://github.com/koalaman/shellcheck/wiki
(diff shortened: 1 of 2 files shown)


skovhus:
Thanks! Seems like you just need to run linting/formatting.

PR #1387, opened 2026-07-03
feat: Add `bashIde.shfmt.additionalArguments` LSP setting.

### LINKED ISSUE #1385: Allow multi-argument commands to start up `shfmt`
### What is the problem this feature will solve?

I have a wrapper executable around `shfmt` that requires multiple arguments to be passed to it, e.g. `/path/to/formatter --shfmt`. To my understanding, the current LSP setting only accepts a string and not a list of strings for running the formatter, so the command will probably fail with `Command '/path/to/formatter --shfmt' not found` or something like that.

### What is the feature you are proposing to solve the problem?

It would be nice if the LSP setting accepted a list of strings (with the current string type getting "promoted" to a singleton list), so I could have my setting be: `[ "/path/to/formatter", "--shfmt" ]`. Then in the language server, `--` could be appended first, before all of the other `shfmt`-specific flags that are controlled by the other LSP settings.

I haven't tested this, but I imagine if the user provided `--` themselves, then maybe they could provide their own set of `shfmt` flags themselves? And so the language server wouldn't need to create a new option for every format option that exists.

### What alternatives have you considered?

Running a regular version of `shfmt`, but the version may differ from what's in my wrapper script.

kylechui:
This will allow a user to extend the arguments being passed to the `shfmt` binary, and resolve #1385.

NOTE: I think that some of the tests had their number of arguments increased by one, since my wrapper `makeShfmtConfig` will impute `languageDialect: 'auto'` when it used to be omitted.

### DIFF
--- a/server/src/__tests__/config.test.ts
+++ b/server/src/__tests__/config.test.ts
@@ -15,6 +15,7 @@ describe('ConfigSchema', () => {
         "shellcheckExternalSources": true,
         "shellcheckPath": "shellcheck",
         "shfmt": {
+          "additionalArguments": [],
           "binaryNextLine": false,
           "caseIndent": false,
           "funcNextLine": false,
@@ -66,6 +67,7 @@ describe('ConfigSchema', () => {
         "shellcheckExternalSources": true,
         "shellcheckPath": "",
         "shfmt": {
+          "additionalArguments": [],
           "binaryNextLine": true,
           "caseIndent": true,
           "funcNextLine": true,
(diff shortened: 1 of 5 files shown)


kylechui:
The tests are largely passing my local machine, except for a few that rely on the `get-options.sh` script. I think this is because I am running NixOS, which doesn't store those completion files in the Linux FHS. I imagine if the CI for this project has tests that spin up the language server in a virtual machine, then they should pass.

skovhus (approved):

skovhus:
@kylechui do you mind fixing the test that fails on CI?

kylechui:
Ah sorry for the late response, thanks for pushing this through!

PR #1392, opened 2026-08-19
feat: resolve bats `load` as a sourcing command

### LINKED ISSUE #181: Support BATS testing framework with go to definition
Loving the language server!  Thanks for putting it together.

A common framework for e2e testing CLI tools and other command line experiences is [BATS](https://github.com/bats-core/bats-core).  It provides a number of useful tools in writing test cases and more.

In particular, it supports the `load` function, which loads the functions from another file for use in the current file.

The "go to definition" functionality of the language server is not supported when other files are `load`ed into the current one.  This functionality would be great for BATS users!

Please let me know if you need any other context.

Thanks!

kolyshkin:
Fixes #181.

### Problem

[Bats](https://bats-core.readthedocs.io) test files pull in helper files with `load test_helper`, not `source`. Since `load` was not recognized as a sourcing command, none of the helper's symbols were reachable from a `.bats` file — go to definition on a helper function returned nothing.

The only workaround today is `bashIde.includeAllWorkspaceSymbols`, which is a rather blunt instrument: it widens completion and definition across the entire workspace, well beyond what the file actually pulls in.

### Change

`load` is now treated as a sourcing command, but **only in `.bats` files** — `load` is a common enough name for an unrelated command or function elsewhere. Path resolution follows what bats itself does: relative to the directory of the test file, retrying with a `.bash` suffix when the given path is not a file.

```bash
#!/usr/bin/env bats

load test_helper        # -> ./test_helper.bash
load ./test_helper.bash # -> ./test_helper.bash
load /abs/path/helper   # -> /abs/path/helper.bash
```

`bats_load_library` is deliberately left out: it resolves against `BATS_LIB_PATH`, which is not statically known, so it would only produce unresolvable source commands (and, with `enableSourceErrorDiagnostics`, spurious diagnostics).

### Tests

- `sourcing.test.ts`: resolution of the forms above, plus the dynamic/unfinished cases, and a test asserting `load` is *not* treated as sourcing outside `.bats` files.
- `analyzer.test.ts`: an end-to-end `findDeclarationLocations` test going from a `.bats` fixture into its helper — with `includeAllWorkspaceSymbols` off, which is exactly the case that did not work before.
- New fixtures under `testing/fixtures/bats/`. The added `.bash` fixture bumps `FIXTURE_FILES_MATCHING_GLOB` and adds one empty entry to two workspace-wide rename snapshots.

Note: `formatter › should format when shfmt is present` fails on my machine both before and after this change — the expected-message regex does not match shfmt 3.13.1's wording. Unrelated to this PR.

### DIFF
--- a/server/src/__tests__/__snapshots__/server.test.ts.snap
+++ b/server/src/__tests__/__snapshots__/server.test.ts.snap
@@ -1248,6 +1248,7 @@ exports[`server onRenameRequest Workspace-wide rename returns correct WorkspaceE
 exports[`server onRenameRequest Workspace-wide rename returns correct WorkspaceEdits for unsourced symbols when includeAllWorkspaceSymbols is true 1`] = `
 {
   "changes": {
+    "file://__REPO_ROOT_FOLDER__/testing/fixtures/bats/test_helper.bash": [],
     "file://__REPO_ROOT_FOLDER__/testing/fixtures/comment-doc-on-hover.sh": [],
     "file://__REPO_ROOT_FOLDER__/testing/fixtures/extension.inc": [],
     "file://__REPO_ROOT_FOLDER__/testing/fixtures/install.sh": [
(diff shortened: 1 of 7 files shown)


skovhus (approved):

kolyshkin:
@skovhus I think you need to enable CI here (there should be a button). If not possible, close and reopen this PR does the trick.

kolyshkin:
rebased on top of merged #1393; CI started!

kolyshkin:
CI is green now but auto-merge is now disabled because of rebase :( PTAL @skovhus

PR #1396, opened 2026-09-09
fix: prevent out of memory crash on cyclic symbolic links

### LINKED ISSUE #1235: OOM error from cyclic symlinks in repo
### Code editor

Emacs

### Platform

Linux (NixOS unstable)

### Version

5.4.0

Open any `.sh` file inside of e.g. https://github.com/fwupd/fwupd/ which contains a `/sys` style directory inside of `src/tests/sys` with cyclic symlinks.

Reproduces perfectly.

Probably needs some kind of cycle detection for symlinks when walking the file system. An example might be the Rust walkdir crate https://github.com/BurntSushi/walkdir/blob/2.5.0/src/lib.rs#L973

### What do you see instead?

This is the stderr after the process exits:

```
<--- Last few GCs --->

[99443:0x42a319f0]   243446 ms: Scavenge 3892.5 (4127.2) -> 3881.1 (4127.9) MB, 11.21 / 0.00 ms  (average mu = 0.141, current mu = 0.014) task; 
[99443:0x42a319f0]   243489 ms: Scavenge 3893.4 (4127.9) -> 3881.9 (4128.7) MB, 10.31 / 0.00 ms  (average mu = 0.141, current mu = 0.014) task; 
[99443:0x42a319f0]   246551 ms: Mark-Compact 3893.9 (4128.7) -> 3879.0 (4129.4) MB, 3017.90 / 0.00 ms  (average mu = 0.128, current mu = 0.115) task; scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0xab5100 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [/run/current-system/sw/bin/bash-language-server]
 2: 0xe7cec0 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [/run/current-system/sw/bin/bash-language-server]
 3: 0xe7d2a4 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [/run/current-system/sw/bin/bash-language-server]
 4: 0x10ac967  [/run/current-system/sw/bin/bash-language-server]
 5: 0x10ac9f9  [/run/current-system/sw/bin/bash-language-server]
 6: 0x10c4590 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::internal::GarbageCollectionReason, char const*) [/run/current-system/sw/bin/bash-language-server]
 7: 0x10c50c7 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/run/current-system/sw/bin/bash-language-server]
 8: 0x1120844 v8::internal::MinorGCJob::Task::RunInternal() [/run/current-system/sw/bin/bash-language-server]
 9: 0xcb1755 node::PerIsolatePlatformData::RunForegroundTask(std::unique_ptr<v8::Task, std::default_delete<v8::Task> >) [/run/current-system/sw/bin/bash-language-server]
10: 0xcb3f42 node::PerIsolatePlatformData::FlushForegroundTasksInternal() [/run/current-system/sw/bin/bash-language-server]
11: 0x7f2880e2b193  [/nix/store/p9rzfyn249qip445ja9xb8grg88jp5br-libuv-1.48.0/lib/libuv.so.1]
12: 0x7f2880e3f892  [/nix/store/p9rzfyn249qip445ja9xb8grg88jp5br-libuv-1.48.0/lib/libuv.so.1]
13: 0x7f2880e2beb0 uv_run [/nix/store/p9rzfyn249qip445ja9xb8grg88jp5br-libuv-1.48.0/lib/libuv.so.1]
14: 0xb2bbb3 node::SpinEventLoopInternal(node::Environment*) [/run/current-system/sw/bin/bash-language-server]
15: 0xc83233 node::NodeMainInstance::Run(node::ExitCode*, node::Environment*) [/run/current-system/sw/bin/bash-language-server]
16: 0xc8361e node::NodeMainInstance::Run() [/run/current-system/sw/bin/bash-language-server]
17: 0xbe0212 node::Start(int, char**) [/run/current-system/sw/bin/bash-language-server]
18: 0x7f287da3127e  [/nix/store/pacbfvpzqz2mksby36awvbcn051zcji3-glibc-2.40-36/lib/libc.so.6]
19: 0x7f287da31339 __libc_start_main [/nix/store/pacbfvpzqz2mksby36awvbcn051zcji3-glibc-2.40-36/lib/libc.so.6]
20: 0xaf5d95 _start [/run/current-system/sw/bin/bash-language-server]

Process bash-ls stderr finished
```

### Additional information

_No response_

superphosphate:
### Summary

`fast-glob` follows symbolic links while walking the workspace (`followSymbolicLinks: true`). A repository that contains a cyclic symbolic link (for example `fwupd`'s `src/tests/sys`) therefore makes the server walk the same directory over and over again until Node runs out of memory:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

This adds a `fast-glob` file system adapter that records the real path of every directory that is read and returns an empty directory when the same real path is encountered again. Symbolic links keep working, cycles are broken.

### Related issues

fixes #1235

### What changed

- `server/src/util/fs.ts`: `getFilePaths` now passes a cycle-safe `fs` adapter to `fast-glob`.
- `server/src/util/__tests__/fs.test.ts`: new tests for glob matching, following symbolic links, cyclic symbolic links and the `maxItems` limit.

### Verification

- `pnpm exec tsc -b` passes.
- `pnpm exec jest --runInBand server/src/util/__tests__/fs.test.ts`: 4/4 pass.
- Without the fix, the `does not follow cyclic symbolic links` test fails with paths such as `loop/loop/loop/.../script.sh`, which is exactly the runaway behaviour from the issue.
- `pnpm exec eslint server/src/util/fs.ts server/src/util/__tests__/fs.test.ts` passes.
- Full suite: 142 passed (4 more than the 138 passed on a clean checkout). The 7 failing suites (`server.test.ts`, `analyzer.test.ts`, `sourcing`, `sh`, `shfmt`, `executables`, `shellcheck`) fail identically on a clean checkout in this environment, so they are unrelated to this change.

### Notes

- Directories reachable through more than one link are now read only once, which also avoids doing the same work twice.
- The test creates the cycle with a junction on Windows (no administrator rights required) and a directory symlink on other platforms.

### DIFF
--- a/server/src/util/__tests__/fs.test.ts
+++ b/server/src/util/__tests__/fs.test.ts
@@ -1,0 +1,93 @@
+import * as fs from 'node:fs'
+import * as os from 'node:os'
+import * as path from 'node:path'
+
+import { getFilePaths } from '../fs'
+
+const symlinkType = process.platform === 'win32' ? 'junction' : 'dir'
+
+const relativePaths = (filePaths: string[], rootPath: string): string[] =>
+  filePaths.map((filePath) => path.relative(rootPath, filePath).split(path.sep).join('/'))
+
+describe('getFilePaths', () => {
+  let rootPath: string
+
+  beforeEach(() => {
+    rootPath = fs.mkdtempSync(path.join(os.tmpdir(), 'bash-language-server-fs-'))
+  })
+
+  afterEach(() => {
+    fs.rmSync(rootPath, { recursive: true, force: true })
+  })
+
+  it('returns the paths of files matching the glob pattern', async () => {
+    fs.writeFileSync(path.join(rootPath, 'script.sh'), '')
+    fs.mkdirSync(path.join(rootPath, 'nested'))
+    fs.writeFileSync(path.join(rootPath, 'nested', 'nested.sh'), '')
+
+    const filePaths = await getFilePaths({
+      globPattern: '**/*.sh',
+      rootPath,
+      maxItems: 100,
+    })
+
+    expect(relativePaths(filePaths, rootPath).sort()).toEqual([
+      'nested/nested.sh',
+      'script.sh',
+    ])
+  })
+
+  it('follows symbolic links to directories', async () => {
+    const targetPath = fs.mkdtempSync(
+      path.join(os.tmpdir(), 'bash-language-server-fs-target-'),
+    )
+
+    try {
+      fs.writeFileSync(path.join(targetPath, 'linked.sh'), '')
+      fs.symlinkSync(targetPath, path.join(rootPath, 'link'), symlinkType)
+
+      const filePaths = await getFilePaths({
+        globPattern: '**/*.sh',
+        rootPath,
+        maxItems: 100,
+      })
+
+      expect(relativePaths(filePaths, rootPath)).toEqual(['link/linked.sh'])
+    } finally {
+      fs.rmSync(targetPath, { recursive: true, force: true })
+    }
+  })
+
+  it('does not follow cyclic symbolic links', async () => {
+    fs.writeFileSync(path.join(rootPath, 'script.sh'), '')
+
+    const loopPath = path.join(rootPath, 'loop')
+    fs.symlinkSync(rootPath, loopPath, symlinkType)
+
+    try {
+      const filePaths = await getFilePaths({
+        globPattern: '**/*.sh',
+        rootPath,
+        maxItems: 100,
+      })
+
+      expect(relativePaths(filePaths, rootPath)).toEqual(['script.sh'])
+    } finally {
+      fs.unlinkSync(loopPath)
+    }
+  })
+
+  it('stops after the maximum number of files', async () => {
+    for (let i = 0; i < 10; i++) {
+      fs.writeFileSync(path.join(rootPath, `script-${i}.sh`), '')
+    }
+
+    const filePaths = await getFilePaths({
+      globPattern: '**/*.sh',
+      rootPath,
+      maxItems: 3,
+    })
+
+    expect(filePaths).toHaveLength(3)
+  })
+})
--- a/server/src/util/__tests__/fs.test.ts
+++ b/server/src/util/__tests__/fs.test.ts
@@ -58,6 +58,31 @@ describe('getFilePaths', () => {
     }
   })
 
+  it('follows several symbolic links to the same directory', async () => {
+    const targetPath = fs.mkdtempSync(
+      path.join(os.tmpdir(), 'bash-language-server-fs-target-'),
+    )
+
+    try {
+      fs.writeFileSync(path.join(targetPath, 'shared.sh'), '')
+      fs.symlinkSync(targetPath, path.join(rootPath, 'a'), symlinkType)
+      fs.symlinkSync(targetPath, path.join(rootPath, 'b'), symlinkType)
+
+      const filePaths = await getFilePaths({
+        globPattern: '{a,b}/**/*.sh',
+        rootPath,
+        maxItems: 100,
+      })
+
+      expect(relativePaths(filePaths, rootPath).sort()).toEqual([
+        'a/shared.sh',
+        'b/shared.sh',
+      ])
+    } finally {
+      fs.rmSync(targetPath, { recursive: true, force: true })
+    }
+  })
+
   it('does not follow cyclic symbolic links', async () => {
     fs.writeFileSync(path.join(rootPath, 'script.sh'), '')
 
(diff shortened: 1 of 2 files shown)


superphosphate:
Addressed the review comment in e6e9f64.

The adapter now skips a directory only when its real path equals the real path of one of its **ancestors**, instead of skipping every already-read real path. That still breaks cycles (the walk can only descend, so an infinite walk has to revisit an ancestor) but no longer drops matches for the same directory reachable through several distinct symbolic links, for example `{a,b}/**` when both `a` and `b` point at the same target.

Added two regression tests:
- `follows several symbolic links to the same directory` (the `{a,b}/**` case from the review)
- `does not follow cyclic symbolic links to an ancestor`

Verification: `tsc -b` passes, 6/6 `fs` tests pass, `eslint` passes.

skovhus:
Feels odd that we need to do a fix on our side, did you investigate if there are any fast-glob issues related to this or workarounds? Also fine to switch library if there are better options now.

superphosphate:
@skovhus good question — I checked before hand-rolling the adapter.

- fast-glob follows symlinks by default and, like node-glob, recurses forever on a cyclic symlink. Upstream treats this as a known limitation with no fix, and the suggested workarounds are `followSymbolicLinks: false` (drops symlink support — this report came from NixOS) or a `deep` cap (an arbitrary cutoff that truncates genuinely deep trees): https://github.com/mrmlnc/fast-glob/issues/74.
- Switching library didn't look clearly better: `tinyglobby` is built on `fdir`, which has its own history of symlink-loop bugs (https://github.com/thecodrr/fdir/issues/125, https://github.com/thecodrr/fdir/issues/135), and Node's built-in `fs.glob` requires Node 22 while this repo supports Node >= 20.

So the adapter keeps `followSymbolicLinks: true` and only skips a directory whose real path matches one of its ancestors — the minimal condition that breaks a cycle, since the walk only descends. I've added a pointer to the fast-glob issue in the code comment (ade07f6).

If you'd prefer migrating to `tinyglobby` (or the built-in `fs.glob` once Node 20 is dropped), I'm happy to open a follow-up PR for that instead.

superphosphate:
The verify (22.x) failure was mine, sorry — fixed in dd86458.

**What happened**

The first version called fs.realpath for *every* directory it read. The workspace-wide rename test initialises the server with the repository root four times, and pnpm install puts a symlink-heavy 
ode_modules there, so the glob hits ~6400 directories. Measured on the repo root:

| | time | 
eaddir calls |
| --- | --- | --- |
| plain fast-glob | 0.5 s | — |
| first version of the patch | 4.2 s | 6366 |

Four of those scans plus the rest of the test crossed the 5 s jest timeout on Node 22 (Node 20 passed because it was just under).

**Fix**

A cycle can only be entered through a symbolic link (the walk only descends, so a plain directory can never point back to an ancestor). The adapter now records which entries are symbolic links from the dirents it already receives and only checks those, comparing 
ealpathSync.native results against the memoized real paths of their ancestors. Directories that are not symbolic links cost a single set lookup.

On the same repository root: **0.52 s plain vs 0.84 s with the adapter**, of which the remaining difference is 1535 
ealpath calls for the ~900 pnpm symlinks (down from 6366 async calls). On Linux the native 
ealpath is a single cheap syscall, so this is a few milliseconds per scan.

**Verification**

- the 6s tests pass.
- Full suite locally: 144 passed (138 on a clean checkout), same 71 pre-existing Windows-only failures as before the change.
- Test list is unchanged: symlinks are still followed, several symlinks to the same directory still both match, cyclic links (including links to an ancestor) are cut.

superphosphate:
Note for whoever reviews: the workflow run for dd86458 is waiting for approval (`action_required`), so the verify jobs have not run on the fix yet. Could you approve/re-run it when you get a chance?

Local check of the test that timed out (`returns correct WorkspaceEdits for sourced symbols`, which initialises the server with the repo root four times):

| | test time (Windows, Node 24) |
| --- | --- |
| before this PR | 8.06 s |
| this PR, dd86458 | 8.67 s |

On Linux the remaining cost is smaller still: the adapter now does one `realpathSync.native` per symbolic link (~900 in a pnpm install) instead of one async `realpath` per directory (~6400), and it does no filesystem call at all for directories that are not symbolic links.

skovhus:
@codex review

skovhus:
@codex review

PR #1422, opened 2026-09-16
fix: resolve forward function definitions from function bodies

### LINKED ISSUE #1213: Cannot jump to function definition when it's defined later in the file
### Code editor

Emacs

### Platform

Linux

### Version

5.4.2

Example code, see the comment:

    #!/usr/bin/env bash

    function foo {
        echo in foo
        bar # cannot jump to definition of bar() from here
    }

    function bar {
        echo in bar
    }

    foo

every time, yes

Editor pointer should go to `function bar {`.

### What do you see instead?

it doesn't

### Additional information

_No response_

skovhus:
Go to definition now resolves a function called from another function before its declaration later in the file. Bash permits this when the call executes after both functions have been defined; previously the language server discarded every later declaration.

Keep the existing behavior for the latest preceding declaration and for top-level calls. Forward lookup is limited to command names and file-scope declarations, so variables and later nested functions do not resolve as forward calls.

Fixes #1213.

Validation: two regressions fail before the fix; analyzer and forward-definition suites pass (42 tests, 23 snapshots); TypeScript compilation and targeted ESLint pass. Independently reviewed by another agent.


skovhus:
@codex review

Issue #1390, opened 2026-07-24
CPU and Memory usage balloons on home directory


Teles-sd:
### Code editor

vim 9.2.0735-1

### Platform

Arch Linux x86_64. Kernel: Linux 7.1.3-arch1-3

### Version

bash-language-server version: `5.6.0`. Client: vim plugin [yegappan/lsp](https://github.com/yegappan/lsp) (no version tag, using [most recent commit](https://github.com/yegappan/lsp/commit/989016ae2ae4cbf304a9ca29478f47fec794493f) as of 24/jul/26)

### What steps will reproduce the bug?

I have a language client set up with the vim plugin [yegappan/lsp](https://github.com/yegappan/lsp), with the following configuration (copied only what I thought was relevant for this):

`~/.vimrc`
```vim
let lspOpts = #{
            \   autoHighlightDiags: v:true,
            \   showDiagOnStatusLine: v:true,
            \   showDiagWithSign: v:true,
            \   showDiagWithVirtualText: v:true,
            \   diagVirtualTextAlign: 'after',
            \ }
autocmd User LspSetup call LspOptionsSet(lspOpts)
let lspServers = [
            \       #{name: 'bashls',
            \           filetype: 'sh',
            \           path: 'bash-language-server',
            \           args: ['start'],
            \       },
            \    ]
autocmd User LspSetup call LspAddServer(lspServers)
```

Then, open a `sh` file at home directory. That's it.

### How often does it reproduce? Is there a required condition?

Consistently. In all these cases:

- Open any `sh` type files at home (e.g. with `vim .bashrc`);
- or files in other directories, but read from home (e.g. with `vim ~/Project/whatever/somefile.sh`);
- or even files at home, but read from other directories (e.g. with `cd ~/Projects && vim ../.bashrc`).

### What is the expected behavior?

When reading some file in, and from, another directory, it behaves ok (e.g. with `cd ~/Projects/whatever/ && vim somefile.sh`).

For comparison, doing as such on a directory and opening one, or a bunch, of `sh` files on vim will have the server memory usage **fluctuate around 113M**.

### What do you see instead?

CPU usage for whatever core is running the server goes to **100%** (ish) and Memory usage keeps going up until it starts to lag everything.

Here's a screenshot of [`btop`](https://github.com/aristocratos/btop) for one such process:

<img src="https://i.imgur.com/0Qp1m0R.png" alt="screenshot of btop showing language server (node) process" style="width: 70%; height: auto;" />

[...]

PR #1447, opened 2026-09-16
fix: bound and cancel background file discovery

skovhus:
Background analysis limits matching files, so a large workspace with few shell scripts can still traverse every directory before finding 500 matches. This bounds discovery at 10,000 directory reads and gives the whole background pass a built-in 10-second elapsed-time budget, covering discovery, pending file reads, and checks between parses. Synchronous analysis already in progress finishes before the deadline can be enforced.

The default glob now prunes hidden entries before scheduling directory reads or symlink lookups. Custom globs retain support for deliberately scanning hidden directories. `backgroundAnalysisIgnore` is the only new user setting, defaulting to `**/node_modules/**` and `**/.git/**`; the existing file-count and glob settings remain available. Budget warnings explain that workspace symbols may be incomplete and suggest exclusions or a narrower glob.

Configuration changes and shutdown cancel pending work. Replacement scans evict background-only documents absent from the new results and explicitly free their syntax trees, fixing stale symbols after an ignore change. Opened/on-demand documents and their sourced dependencies remain cached. Background reads cannot overwrite an opened document's unsaved contents, including when it is opened while a read is pending.

Related to #1390. This addresses reproduced sparse-workspace traversal and unnecessary cache traversal; the reporter's complete home-directory failure has not been reproduced, so this does not claim to close that report.

Validation:

- Actual CLI initialized in a large home directory: the revised defaults discover 147 files and analyze 146, including 86 discovered work scripts, in 0.82–0.91 seconds across two local warm runs. Before pruning, #1447 analyzed 26 files and discovered 18 work scripts. The directory cap remains unchanged. Sampled peak RSS was 320–382 MiB for the revised runs; these are individual measurements, not a controlled benchmark.
- Synthetic workspace with 24,002 directories and one shell file confirms the 10,000-directory cap independently of matching-file count.
- Regression coverage includes ignore-change eviction and tree cleanup, preserving on-demand and sourced documents, removing dependencies no longer sourced after a reread, protecting newer scans from late completion, unsaved edits during pending reads, default hidden-directory pruning/custom hidden globs, one deadline across discovery and reads, synchronous parsing delaying timers, cancellation, and late callbacks. The ignore-eviction regression fails against the previous PR head and passes with this fix.
- TypeScript build and full ESLint pass. Full Jest suite: 345 tests and 90 snapshots pass; 2 existing tests skipped.
- Independently reviewed, including cached-tree ownership and cancellation races.


skovhus:
@codex review

